PR #15, opened 2017-06-22
fixed Glitch link

Utopiah:
Was missing trailing /


Utopiah:
Eh... I don't get it, it stil doesn't work, I'll open a bug report on glitch.

PR #25, opened 2017-07-11
Toward 0.3.0

machenmusik:
a few bugfixes: networked-share, Firebase+WebRTC;
expose NAF options for convenience;
update getDirtyComponents to version in networked.js
(which allows selector-based schema components to work,
which is needed for template-based physics)


haydenjameslee:
Lgtm

PR #28, opened 2017-07-12
Toward 0.3.0

machenmusik:
add networked-share to getNetworkOwner
make remote template instantiations monkey-patch the root element
support falsy selector to mean this.el
(networked-share) automate default networkId and owner
add my quick-n-dirty ball-dropping example (plaid-polo)
add useShare option, to use networked-share for remote entities
add ball-room and ball-room-4 examples


haydenjameslee:
Gave it a brief read through. Lgtm.

PR #40, opened 2017-09-28
Fix EasyRtcAdapter send/broadcastData to use WebRTC when possible.

robertlong:
EasyRtcAdapter's sendData method was using [easyrtc.sendPeerMessage](https://easyrtc.com/docs/client-api/Easyrtc.php#sendPeerMessage) which doesn't use WebRTC data channels when they are connected. The broadcastData method was also using WebSockets via [easyrtc.sendWS](https://easyrtc.com/docs/client-api/Easyrtc.php#sendWS).

This fix makes the EasyRtcAdapter's sendData and broadcastData methods use [easyrtc.sendData](https://easyrtc.com/docs/client-api/Easyrtc.php#sendData) which uses WebRTC when available and WebSockets as a fallback.


haydenjameslee:
Great catch! Merging now and will release as 0.3.1 since this is very incorrect behaviour.

PR #46, opened 2017-10-09
Fix firebase app

machenmusik:
The Firebase adapter was not using the specified app name (which becomes appId).  Instead, the default app name was always used, which could cause conflicts if actually trying to use more than one app per Firebase endpoint.

On the assumption that said behavior was unintentional, these changes correct that behavior.

(Note: the sanitization of the selector portion of the Firebase key was necessary in prior versions, but may have been replaced by different mechanisms in the latest, so not sure if that portion is still required.)

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -52,16 +52,8 @@ module.exports.createNetworkId = function() {
 
 module.exports.delimiter = '---';
 
-function sanitizeKey(key) {
-  return key.replace(/\./g, '!').replace(/\#/g, '@').replace(/\[/g, '{').replace(/\]/g, '}'); 
-}
-
-function desanitizeKey(key) {
-  return key.replace(/\!/g, '.').replace(/\@/g, '#').replace(/\{/g, '[').replace(/\}/g, ']'); 
-}
-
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -52,8 +52,16 @@ module.exports.createNetworkId = function() {
...
+  return key.replace(/\./g, '!').replace(/\#/g, '@').replace(/\[/g, '{').replace(/\]/g, '}'); 
+}
+
+function desanitizeKey(key) {
+  return key.replace(/\!/g, '.').replace(/\@/g, '#').replace(/\{/g, '[').replace(/\}/g, ']'); 
+}
+
 module.exports.childSchemaToKey = function(schema) {
-  var key = (schema.selector || '')
+  var key = sanitizeKey(schema.selector || '')
```
haydenjameslee: @machenmusik this shouldn't need to here due to this line: https://github.com/haydenjameslee/networked-aframe/blob/master/src/adapters/FirebaseWebRtcAdapter.js#L175

Are there cases that this doesn't solve? It would be best if the firebase restrictions didn't leak out into all keys.

PR #47, opened 2017-10-16
Fix Firebase adapter

haydenjameslee:


haydenjameslee:
Based on https://github.com/haydenjameslee/networked-aframe/pull/46

Issue #3, opened 2017-03-16
Questions to use it in Augmented Reality


jeromeetienne:
I love what you did, the easyness of it is so nice!

i would love to use that in augmented reality (with [ar.js](https://github.com/jeromeetienne/ar.js)) but for that i need to change the usual behavior a bit, so i have 2 questions :)

1. i would like to get the avatar, not attached to the camera, but somewhere above the marker. Where should i look to see how to change avatar position.
2. i would like to move the avatar according to controls on a phone, so some kind of touchscreen joystick. Where should i look to be able to controls the position of the avatar from javascript ?

Thanks :)

haydenjameslee:
Thanks!

@jeromeetienne ar.js looked great. It'd be awesome to have an official example that shows NAF can work well for AR.

If i'm understanding you right you're wanting an avatar that is completely independent of the camera. That's totally possible, you'll just think of the object as a regular network entity, and then enable the controls for the local user.

Something like this should work:

```javascript
var avatar = NAF.entities.createNetworkEntity("ar-avatar", position, rotation);

avatar.SetAttribute("gamepad-controls","lookEnabled: false");
```

and then include donmccurdy's gamepad-controls component: https://github.com/donmccurdy/aframe-gamepad-controls

Let me know if that works!

jeromeetienne:
excelent! i love the idea of the official example. i will try and keep you updated :)

haydenjameslee:
Closing this for now. Feel free to post further questions here though!

Issue #2, opened 2017-03-12
Mobile Clients not visible


karunakaruna:
Potentially when using HTTPS for signallingURL.

Desktop clients show:
Successfully started datachannel to  Ao5IrnsQKoP2niV5
Opened data channel from Ao5IrnsQKoP2niV5
...
Closed data channel from Ao5IrnsQKoP2niV5

Mobile clients show only:
Closed data channel from Ao5IrnsQKoP2niV5

haydenjameslee:
This was using the example demos (from private convo).

Would be good to repro locally using localhost. I'll try in the next day or two, if somebody else wants to try go ahead. You'll need to change the signalling URL to the local IP of your PC and then connect directly to that IP with your phone.

haydenjameslee:
I tried to repro just now. It was working fine for me on mobile and using https://glitch.com/~networked-aframe to test. There may have been one time when this error occurred but I can't confirm that.

kfarr:
To provide a device test data point, I wasn't able to get local or remote demos working with iOS 10.3.2 Beta 4 Safari. I can get both demos working open in separate tabs on desktop PC Chrome. I was able to get Android Chrome (Samsung Galaxy S7 Chrome, not webvr) browser to broadcast avatar position to the desktop during one session, but took a few tries on both the desktop and mobile browser to get this to work.

Does iOS support webrtc at all? Looks like not: http://iswebrtcreadyyet.com/

For remote demo I'm using [glitch.io demo](https://networked-aframe.glitch.me/). For local demos I am [following this tutorial](https://github.com/haydenjameslee/networked-aframe/blob/master/docs/Tutorial:%20Create%20your%20first%20Networked-Aframe%20experience.md).

haydenjameslee:
This is most likely due to webrtc not being supported on iOS.

I'll be making a switch in an upcoming release which lets you use socket.io for all networking, instead of WebRTC. This will work on iOS and avoid any of the problems with establishing P2P connections.

brendanluu:
Yes! iOS support with socket.io would be very useful

haydenjameslee:
@brendanluu Cool i'll probably just hack together a quick option and add it in the next release (which also features hand controller support)

haydenjameslee:
@brendanluu in NAF 0.2.0 all messages get sent across socket.io by default. WebRTC is now optional.

Vamoss:
[...]

PR #49, opened 2017-10-21
Allow registering network adapters via the NAF global

robertlong:
Currently all network adapters are included in the networked-aframe library. Eventually it would make sense if all adapters were split out of the core networked-aframe library and into separate modules.

If you are including the easyrtc adapter via script tags it might look like this:
```
<script src="https://aframe.io/releases/0.7.0/aframe.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/socket.io/1.4.5/socket.io.min.js"></script>
<script src="https://unpkg.com/networked-aframe/dist/networked-aframe.min.js"></script>
<script src="https://unpkg.com/networked-aframe/dist/networked-aframe-easyrtc-adapter.min.js"></script>
```

Inside each adapter you would call `NAF.adapters.register(adapterName, AdapterClass);`

So the easyrtc adapter would look like this:

```
class EasyRTCAdapter {
  // ...
}

NAF.adapters.register("easyrtc", EasyRTCAdapter);
```

Additionally this exposes an API for developers to create third party network adapters. The Mozilla MR team has been working on one that uses the Janus WebRTC server as a backend:
https://github.com/netpro2k/networked-aframe/blob/feature/janus-adapter/src/adapters/JanusAdapter.js

This PR includes the `NAF.adapters.register()` API, but does not yet split out the existing adapters into separate modules. I made each of the adapters ready to be split out into their own modules by removing the INetworkAdapter class and other references to networked-aframe. All adapters now reference networked-aframe through the NAF global. The next steps to breaking out these adapters would be to remove them from the AdapterFactory, update the build scripts, and bump the major version.

In this PR there should be no breaking changes to existing apps. Although, the basic-uws.html and deepstream-basic.html examples were broken on master for me and I couldn't test them. I tested all of the other examples successfully.


haydenjameslee:
Sounds great and makes total sense. Its probably worth finishing the remaining tasks before merging right?

robertlong:
If you merge right now, there will be no breaking changes. The remaining work would split out all of the adapters into separate bundles which would need to be included separately. It's a super simple yet breaking change. Depends how strict you are being with your versioning.

I'm finishing up the JanusAdapter which utilizes this API and want to make sure the NetworkAdapter interface is solid before merging. Once that is done I can make the changes I suggested. Whether that work happens in this PR or a separate PR is up to you.

robertlong:
I've gone ahead and removed the adapters from the AdapterFactory. Now I'm trying to decide what is the best way to build and distribute the adapters. I have 3 possible options I'm looking into and would like some feedback.

1. Keep all of the existing adapters in the current repository. Output the built bundles in `/dist/adapters`.
    Requiring via script tag becomes:
    ```
    <script src="https://unpkg.com/networked-aframe/dist/adapters/naf-easyrtc-adapter.min.js"></script>
    ```
    Via NPM:
    ```
    require("networked-aframe/dist/adapters/naf-easyrtc-adapter");
    ```

2. Split the adapters into separate repositories. Ex. `naf-easyrtc-adapter`. 

3. Use a mono repository via Lerna or Yarn. This is a little more complex, but I think better for things that are dependent on one another. Networked AFrame physics could be broken out into its own package as well.

    Yarn also seems to have made this a bit easier with their workspaces feature:
    https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/

I'm also fine with landing just the first commit of this PR now and taking our time to figure out how to split everything up. It could be a big reorganization of the project and deserves some thought.

haydenjameslee:
I'd like to keep the easyrtc adapter in the library source just for convenience. To store others I'd vote for #2 over #3 just due to familiarity. I think it makes sense to store in different repos that can be updated indpendentently. We may run into issues syncing the version numbers of adapters with NAF if we ever change the NAF adapter interface, however I don't foresee that happening very often.

haydenjameslee:
Maybe unrelated to this PR but as far as adapter interface changes go, I think it makes sense to add the `webrtc` bool back into the `networked-scene` component. Since all adapters that support webrtc signalling will necessarily have a means of transferring data by non-webrtc APIs during the signalling process, they should also be able to do "websocket only" and therefore you should be able to choose webrtc or not with every adapter (or at least every adapter that supports webrtc, some may be websocket only).

Ie. We get rid of the separate `wsEasyRTC` adapter and only have the `easyrtc` adapter, then specifiying webrtc = true / false in `networked-scene` will decide webrtc or not.

robertlong:
Cool, I'll split those adapters out then and move their examples to the separate repositories. [naf-janus-adapter](https://github.com/robertlong/naf-janus-adapter) can be used as a template. I'll leave the `EasyRTCAdapter` in the project as the default.

Let's discuss the second comment in another PR. We don't currently have any plans to send non-signaling data over websockets in the JanusAdapter. Instead we are going to be using reliable datachannels to send that data. WebRTC is currently a requirement, but that may change.

netpro2k:
Rather than having a `webrtc` option I think it would be better to just have a general way of passing arguments to an adapter, every adapters options or going to be different so it doesn't make sense to try and bring them all up to the top level. Questionable if this means `audio` should also get made into an adapter argument (I can see arguments either way)

robertlong:
Ok, I've finished splitting out all of the adapters in to separate repositories. I created a separate organization where these can live and invited you. As soon as you accept I can make you an owner and you can decide whether you want to move the core library over as well.

https://github.com/networked-aframe
https://github.com/networked-aframe/naf-uws-adapter
https://github.com/networked-aframe/naf-firebase-adapter
https://github.com/networked-aframe/naf-deepstream-adapter

Some changes I've made:

Build System (This is what I used for [naf-janus-adapter](https://github.com/robertlong/naf-janus-adapter) and just copied everything over, feel free to keep or change whatever you want):
* upgraded to webpack 3
* setup webpack with webpack-dev-server and webpack-dev-middleware where useful.
* added a Prettier .eslintrc
* used the `babel-preset-env` instead of babel-preset-es2015`
* used `babel-minify-webpack-plugin` instead of uglifyjs through browserify.

Documentation:
* Added heroku deploy button to the uws adapter repo.
* Updated networked-aframe README.md to reference the new repositories.

Fixes:
* Removed lodash as a dependency of firebase-key-encode and fixed a bug in it ([PR](https://github.com/chrisvxd/firebase-key-encode/pull/1)) This removes lodash as a dependency of networked-aframe.

haydenjameslee:
Makes sense to have its own org now that there will be multiple repos. Much better than just referencing them all in the README. I'm happy to move the main repo over too, I'll look into how best to do that.

@netpro2k Agreed some of those `networked-scene` properties have now become adapter specific and we should move to a better system. We can probably do better than the URL query parameter format, but I'm not sure what that would be yet. Definitely the scope of another PR.

@robertlong That Heroku deploy button is awesome! Having similar "Remix on Glitch" buttons would be cool too. The build system sounds good. I don't have much experience configuring JS build systems so can't provide much feedback there.

robertlong:
Yeah. I wanted to do a remix on glitch button for all of them. I’ll try to get that in soon.

haydenjameslee:
Just merged Positional Audio (https://github.com/haydenjameslee/networked-aframe/pull/50) and it caused some conflicts here. It also introduces a new function on the IAdapter interface, getMediaStream (https://github.com/haydenjameslee/networked-aframe/blob/master/src/adapters/INetworkAdapter.js#L22). None of the other adapters support audio yet though so it won't matter.

robertlong:
Fixed the merge conflicts and tested the changes.

haydenjameslee:
Something very odd is happening when I checkout the PR:

image

I've npm prune'd, npm installed and npm updated.

robertlong:
Looks like it was caused by a circular dependency. Should be fixed now.

PR #50, opened 2017-10-21
Add positional audio

### LINKED ISSUE #11: Support positional audio (based on avatar positions)
It'd be nice to spatialise the audio (via `navigator.getUserMedia`) emitted from the avatars in the 3D space.

/cc @caseyyee.

netpro2k:
Adds support for positional audio via a `networked-audio-source` component. Adapters must now expose a `getMediaStream(cleintId)` method that returns a Promise resolving to a WebRTC `MediaStream`. This should also expose a `LocalMediaStream` when given the local client id.

If/when https://github.com/aframevr/aframe/pull/3042 is merged, `networked-audio-source` should be modified to delegate to the `sound` component to handle the actual audio. For now the most barebone version of this has been directly embeded in `netowrked-audio-source`.

The `NAF.connection.adapter.getMediaStream(clientId)` API is useful externally to get access to to the audio data (for providing visual audio feedback for example), but I have not documented it in the README yet. It probably makes sense to think about how we want to expose these sorts of things before suggesting people use them.

closes #11


server/static/basic-audio.html
```diff
@@ -0,0 +1,128 @@
...
+      <a-assets>
+
+        <img id="grid" src="https://img.gs/bbdkhfbzkk/stretch/https://i.imgur.com/25P1geh.png" crossorigin="anonymous">
+        <img id="sky" src="http://i.imgur.com/WqlqEkq.jpg" crossorigin="anonymous" />
+
+        <!-- Templates -->
+
+        <!-- Avatar -->
+        <script id="avatar-template" type="text/html">
+          <a-entity class="avatar" networked-audio-source>
```
haydenjameslee: It could be cool to move the networked-audio-source component onto an entity at the mouth of the avatar

haydenjameslee:
Haven't tested it myself yet but looks good!

server/static/basic-audio.html
netpro2k: I don't think this would make much practical difference so I would rather not make people think it has to be on its own entity or something.

HAZARDU5:
I'm really excited about this! :-D

PR #59, opened 2017-11-22
There is no file called 'server.js' in ./server

MilesMcBain:
Following along with the getting started tutorial npm threw: `Error: Cannot find module '/home/miles/repos/naf-tutorial/server/server.js` when I tried to `npm start`. I think you meant to use the `easyrtc-server.js`?


haydenjameslee:
Nice catch :) Thanks!

PR #64, opened 2017-12-15
Ownership transfer (WIP)

robertlong:
Adds the ability to take ownership of entities to control who is responsible for syncing the entity's component data.

Adds the utility methods `NAF.utils.takeOwnership(entityEl)` and `NAF.utils.isMine(entityEl)`.

Includes a new example `ownership-transfer.html` where users can spawn cubes at their current location with the space bar and then take control of all the cubes in the room with enter. Cubes switch direction when ownership changes and their color changes on first ownership change.

### TODO:
- [x] Update documentation
- [x] Add example link
- [x] In the example make the color change to the owner avatar color.


src/components/networked.js
```diff
@@ -269,6 +280,20 @@ AFRAME.registerComponent('networked', {
...
       entityData = Compressor.decompressSyncData(entityData, this.getAllSyncedComponents());
     }
+
+    if (entityData.lastOwnerTime < this.lastOwnerTime) {
+      return;
+    }
+
+    if (this.data.owner !== entityData.owner) {
+      // TODO: File issue for partial set attribute.
+      // this.el.setAttribute("networked", { owner: entityData.owner });
```
netpro2k: Hmm, really strange that this didn't work, definitely worth investigating. Did the 3 argument version also not work? `this.el.setAttribute("networked", "owner", entityData.owner)`

src/utils.js
```diff
@@ -79,13 +79,26 @@ module.exports.childSchemaEqual = function(a, b) {
...
+
+module.exports.takeOwnership = function(entity) {
+  const networkedEntity = getNetworkedEntity(entity);
+  const networkId = networkedEntity.getAttribute("networked").networkId;
+  // TODO: File issue for partial set attribute.
+  networkedEntity.setAttribute("networked", { owner: NAF.clientId, networkId: networkId });
+};
+
+module.exports.isMine = function(entity) {
+  return getNetworkedEntity(entity).components['networked'].data.owner === NAF.clientId;
```
netpro2k: Should defer to the `isMine` method on the `networked` component.

src/components/networked.js
```diff
@@ -40,19 +40,32 @@ AFRAME.registerComponent('networked', {
...
     if (this.data.owner === '') {
+      this.lastOwnerTime = NAF.utils.now();
+
       this.setNetworkIdWhenConnected();
       // Only send the initial sync if we are connected. Otherwise this gets sent when the dataChannel is opened with each peer.
       // Note that this only works because the reliable messages are sent over a single websocket connection.
       // If they are sent over a different transport this check may need to change
       if (NAF.connection.isConnected()) {
-        this.syncAll();
+        this.waitForTemplate(() => {
```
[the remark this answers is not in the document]
robertlong: Weird. It is in another PR https://github.com/networked-aframe/networked-aframe/pull/63. Idk why it shows up in this diff.

src/components/networked.js
robertlong: Yeah the three argument one failed as well. I'll file a bug in AFrame.

brianpeiris: I think I fixed this in #69

robertlong: Yup, that seems to have fixed it! Nice catch!

src/components/networked.js
```diff
@@ -39,7 +39,11 @@ AFRAME.registerComponent('networked', {
       this.registerEntity(this.data.networkId);
     }
 
+    this.lastOwnerTime = -1;
+
     if (this.data.owner === '') {
+      this.lastOwnerTime = NAF.utils.now();
```
netpro2k: Might be cleaner to set this in `onConnected` when owner gets set.

src/components/networked.js
```diff
@@ -267,6 +287,18 @@ AFRAME.registerComponent('networked', {
     if (entityData[0] == 1) {
       entityData = Compressor.decompressSyncData(entityData, this.getAllSyncedComponents());
     }
+
+    if (entityData.lastOwnerTime < this.lastOwnerTime) {
```
netpro2k: You are checking the opposite condition bellow`this.lastOwnerTime < entityData.lastOwnerTime` so I don't think this check is needed.

src/utils.js
```diff
@@ -65,13 +65,28 @@ module.exports.childSchemaEqual = function(a, b) {
...
     entity = entity.parentNode;
   }
   return entity;
 };
 
+module.exports.getNetworkedEntity = getNetworkedEntity;
+
+module.exports.takeOwnership = function(entity) {
+  const networkedEntity = getNetworkedEntity(entity);
+  if(!networkedEntity) {
```
netpro2k: This should log an error similar to https://github.com/networked-aframe/networked-aframe/blob/master/src/components/networked-audio-source.js#L24

src/components/networked.js
```diff
@@ -54,6 +58,22 @@ AFRAME.registerComponent('networked', {
...
+    if(owner && !this.isMine() && lastOwnerTime < now) {
+      networkedEntity.setAttribute("networked", { owner: NAF.clientId });
+      this.lastOwnerTime = now;
+      return true;
+    }
+    return false;
+  }
+
+  update: function(oldData) {
+    this.syncAll();
```
netpro2k: This feels a bit heavy handed, id imagine this results in sending a bunch more sync packets during init since update will get called as owner and networkId are set. I assume this is being done to handle owner changing?

We had discussed making owner no longer a component property, since its invalid to directly set it, so why not just call `syncAll` inside `takeOwnership`

src/adapters/NoOpNetworkAdapter.js
```diff
@@ -21,6 +21,8 @@ class NoOpNetworkAdapter extends NafInterface {
 
   getMediaStream(clientId) { return Promise.reject("Interface method not implemented: getMediaStream")}
 
+  getTime() { return Date.now(); }
```
netpro2k: I vote we call this something like `getServerTime`. Also the rest of the methods in noop adapter just throw a not implemented error, this should probably do the same.. That said we should probably come up with a nicer way to do this. I forget why we don't just have a base `Adapter` that is not registered but is subleased by all the adapters so that we can provide functionality like this as a base.

server/static/js/toggle-ownership.component.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * Rotate the entity every frame if you are the owner.
+ * When you press enter take ownership of the entity,
+ * spin it in the opposite direction and change its color.
+ */
+AFRAME.registerComponent('toggle-ownership', {
```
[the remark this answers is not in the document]
brianpeiris: The Altspace SDK used an `ownOn` property that took an event name to decouple ownership from the action.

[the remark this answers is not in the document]
brianpeiris: It also emitted ownership gained/lost events.

src/adapters/EasyRtcAdapter.js
```diff
@@ -54,22 +57,46 @@ class EasyRtcAdapter {
     this.easyrtc.setPeerListener(messageListener);
   }
 
+  updateTimeOffset() {
+    const clientSentTime = Date.now();
+
+    return fetch(document.location.href)
```
brianpeiris: Can we fetch a HEAD request instead of a GET request?

brianpeiris: Also not sure how this would react if the page happens to be cached. We may want to force a `no-cache`.

src/adapters/NoOpNetworkAdapter.js
robertlong: I believe we got rid of the network adapters all extending some base adapter because of circular dependencies. Ex.

```

File: src/adapters/EasyRtcAdapter.js [hunk 11]
```diff
@@ -54,22 +57,46 @@ class EasyRtcAdapter {
...
+      .then(res => {
+        var precision = 1000;
+        var serverReceivedTime = new Date(res.headers.get("Date")).getTime() + (precision / 2);
+        var clientReceivedTime = Date.now();
+        var serverTime = serverReceivedTime + ((clientReceivedTime - clientSentTime) / 2);
+        var timeOffset = serverTime - clientReceivedTime;
+        this.timeOffsets.push(timeOffset);
+        this.avgTimeOffset = this.timeOffsets.reduce((acc, offset) => acc += offset, 0) / this.timeOffsets.length;
```
brianpeiris: Wouldn't a running average of the last N offsets be better?

src/adapters/EasyRtcAdapter.js
```diff
@@ -54,22 +57,54 @@ class EasyRtcAdapter {
     this.easyrtc.setPeerListener(messageListener);
   }
 
+  updateTimeOffset() {
+    const clientSentTime = Date.now() + this.avgTimeOffset;
+
+    return fetch(document.location.href, { method: "HEAD", cache: "no-cache" })
```
gfodor: This is a temporary way to get some kind of server clock until we figure out an NTP based solution (or something similar) right?

src/components/networked.js
```diff
@@ -267,6 +273,18 @@ AFRAME.registerComponent('networked', {
     if (entityData[0] == 1) {
       entityData = Compressor.decompressSyncData(entityData, this.getAllSyncedComponents());
     }
+
+    // Avoid updating components if the entity data received did not come from the current owner.
+    if (entityData.lastOwnerTime < this.lastOwnerTime || 
+          (this.lastOwnerTime === entityData.lastOwnerTime && this.data.owner > entityData.owner)) {
```
gfodor: I'm not well versed with this code, but aren't `data.owner` and `entityData.owner` client ids? If so why are they being compared with a `>` operator here and not `!=`?

src/adapters/EasyRtcAdapter.js
netpro2k: It's up to the adapter to decide how it wants to compute time offsets. This is a simple version that will work against any http server for the easyrtc adapter. We are free to do it however we like for our janus adapter.

src/components/networked.js
netpro2k: I belive this is to handle the unlikely case where two people take ownership at the exact same timestamp. In this case the client with the larger id wins. This is just to ensure that everything remains eventually consistent.

haydenjameslee: @gfodor you're right that `data.owner` will be a clientId. The type of clientId is set by the adapter ([passed into the `connectSuccess` callback](https://github.com/netpro2k/networked-aframe/blob/a9107247f9e82145d9fce75fe4ab764791fb02dc/src/NetworkConnection.js#L66) so as long as that type supports the `>` operator this will work as expected. Restricting clientId's to strings and numbers seems reasonable.

haydenjameslee: Actually regardless of all that the Networked component schema restricts the owner to a string.

gfodor:
Yup makes sense, sorry the reason for the consistent tie-breaking should
have been obvious

haydenjameslee:
I just merged https://github.com/networked-aframe/networked-aframe/pull/72 which caused the simple conflicts here.

src/adapters/WsEasyRtcAdapter.js
```diff
@@ -5,9 +5,14 @@ class WsEasyRtcInterface {
     this.app = 'default';
     this.room = 'default';
     this.connectedClients = [];
+
+    this.serverTimeRequests = 0;
```
haydenjameslee: There's starting to be a lot of repeated code between this and EasyRtcAdapter.js. We can worry about this in a separate PR but I think it'd be useful to change back to having a `webrtc` flag in `networked-scene` that you can toggle.

haydenjameslee: [I've created an issue for this](https://github.com/networked-aframe/networked-aframe/issues/80)

robertlong:
I think this PR is ready to be merged @haydenjameslee.

PR #67, opened 2017-12-22
Use setAttribute to change component data

brianpeiris:
Changing component data directly is bad practice as it leads to internal inconsistencies in a-frame. 
For example, the network id may get clobbered if a-frame calls `updateComponents` after an asset timeout.

### DIFF
--- a/src/components/networked.js
+++ b/src/components/networked.js
@@ -131,8 +131,7 @@ AFRAME.registerComponent('networked', {
   },
 
   onConnected: function() {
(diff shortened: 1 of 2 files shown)


haydenjameslee:
I'm hoping this is the solution for another problem we found: https://github.com/networked-aframe/networked-aframe/pull/64/files#r157650229 where the networkId was being reset unless we   set it again. @netpro2k @robertlong

PR #68, opened 2017-12-25
Travis CI config and fixes

brianpeiris:
150 tests is a decent test suite! Let's actually run them regularly.


.eslintrc.json
```diff
@@ -0,0 +1,18 @@
...
+    "rules": {
+      "no-redeclare": "off",
+      "no-unused-vars": "off",
+      "no-undef": "off",
+      "no-mixed-spaces-and-tabs": "off",
+      "no-extra-semi": "off",
+      "no-empty": "off",
+      "no-console": "off"
+    }
+}
```
brianpeiris: Switched from `semistandard` to `eslint` since the codebase violates a lot of semistandard in its current state.
Still had to turn off some eslint rules. I'd say the first four are ones we'd really want to turn back on once the offending code is cleaned up.

.travis.yml
```diff
@@ -0,0 +1,39 @@
...
+    - CI_ACTION="npm run dist"
+
+branches:
+  only:
+    - travis
+
+cache:
+  directories:
+    - node_modules
+
```
brianpeiris: Travis config is mostly a copy of [aframe's config](https://github.com/aframevr/aframe/blob/master/.travis.yml).

package.json
```diff
@@ -35,7 +35,7 @@
     "socket.io-client": "^1.4.5"
   },
   "devDependencies": {
-    "aframe": "0.5.0",
+    "aframe": "0.7.0",
```
brianpeiris: Did this to verify that the tests pass with 0.7.0. Not sure if you intentionally kept it at 0.5.0,

tests/unit/ComponentHelper.test.js
```diff
@@ -71,17 +76,18 @@ suite('ComponentHelper', function () {
...
+      var child = createEntityWithComponents({
         'position': { x: 5, y: 5, z: 1 },
       });
       child.className = 'child';
       el.appendChild(child);
+      scene.appendChild(el);
       var componentsToCheck = ['position', { selector: '.child', component: 'position' }];
 
-      utils.whenEntityLoaded(child, function() {
+      utils.whenEntityLoaded(el, function() {
```
brianpeiris: Running the tests in both Firefox and Chrome (via karma) uncovered a race condition with this suite. Chrome would report that the child entity was not loaded even though the parent entity reports as loaded. Vice versa in Firefox. Something to do with the specifics of the web component implementation/polyfill, I suspect. 

Both browsers behave identically if you force the order of operations by constructing the DOM tree fully, before finally adding it to the scene.

tests/unit/networked_remote.test.js
```diff
@@ -79,7 +79,6 @@ suite('networked_remote', function() {
 
     test('updates root immediately', sinon.test(function() {
       this.stub(component, 'networkUpdate');
-      this.stub(component, 'waitForTemplateAndUpdateChildren');
```
brianpeiris: This function no longer exists. It never needed to be stubbed in the first place, as far as I could tell.

webpack.config.js
```diff
@@ -11,4 +11,4 @@ module.exports = {
             }
         ]
     }
-};
```
brianpeiris: Bah, line endings. I can fix this if you care.

brianpeiris:
Tests are now passing in Travis on this branch: https://travis-ci.org/brianpeiris/networked-aframe/builds/321230682

package.json
haydenjameslee: This was just outdated, it wasn't intentional

haydenjameslee:
Thanks for this! Have been meaning to set it up for a while.

.eslintrc.json
haydenjameslee: Cleaned the code up and turned those switches on: https://github.com/networked-aframe/networked-aframe/pull/91

Issue #80, opened 2018-01-08
Idea: Merge wsEasyRTC adapter with EasyRTC adapter and add webrtc bool to networked-scene


haydenjameslee:
Any WebSocket-only adapter can theoretically do WebRTC without much modification. Once there is a WebSocket connection all the signaling messages can transfer across it and then use WebRTCPeer.js like the [naf-firebase-adapter](https://github.com/networked-aframe/naf-firebase-adapter/tree/master/src) to implement WebRTC.

I'm proposing we add another field to the `networked-scene` schema that allows you to toggle between WebSockets and WebRTC, thus halving the number of adapters we need and making it easier to test different architectures very quickly.

Thoughts?

haydenjameslee:
Ok I made this example real quick: https://github.com/networked-aframe/networked-aframe/pull/81

I don't like how messy it makes [EasyRTCAdapter.js](https://github.com/networked-aframe/networked-aframe/pull/81/files#diff-15d3ea80b33e0826ffcbb0a8e03af7e0).

Maybe this can be cleaned up a lot, but this is probably one of the main reasons I opted for separating the WebSocket adapters into their own files last year...

haydenjameslee:
As far as the API goes, an alternative would be `networked-scene="transport: websocket / webrtc"`

haydenjameslee:
@netpro2k That logic makes sense, I like the idea of `adapterOptions`. I can make a PR for it but if your PR already does some of it I'd be interested to see it. Feel free to make the PR.

hthetiot:
Should be fixed by #175 via Open-EasyRTC latest WebRTCAdapter.js update

vincentfretin:
I know that the janus adapter and now the dialog/mediasoup adapter used by Hubs doesn't use the `audio: bool` param in `networked-scene`, but use `NAF.connection.adapter.enableMicrophone(bool)` to enable an audio track
and @kylebakerio was bite by this because it's not the same API.

kylebakerio:
Going forward, setting options should definitely be unified. While in general I like the idea of just setting whatever adapter options are available in the networked-scene component properties, I _really_ appreciated the simplicity of supplying my chosen stream to janus within the JS, even if it did feel a bit confusing of an API. 

[...]

PR #81, opened 2018-01-08
Example implementation of webrtc toggle in networked-scene

haydenjameslee:
Discussion: https://github.com/networked-aframe/networked-aframe/issues/80

This is just a reference implementation, it can probably be improved.


haydenjameslee:
Closing this, we're gonna take a different direction as per https://github.com/networked-aframe/networked-aframe/issues/80

PR #85, opened 2018-01-12
use almost-equal for positions

arpu:


ngokevin:
I think there's a case that pose should just always sync and save time on checking/diffing.

haydenjameslee:
Yeah that'd make sense for a lot of entities. I could see some use cases where the position changes very rarely, like pieces in a game of checkers. Maybe we add a global option to always sync pose and have it default on. It could also be an option on the entity itself, which would probably make more sense.

haydenjameslee:
Closing until the above issues are addressed.

PR #93, opened 2018-02-07
Refactor networked templates

robertlong:
Currently networked templates use the `aframe-template-component` which appends the template elements to the networked object. This makes it impossible for us to make templates that have components on the root of the networked entity.

Example:
```
<script id="avatar-template" type="text/html">
  <a-entity gltf-model="#avatar-model">
    <a-text position="0 1 0" value="Avatar Name" ></a-text>
  </a-entity>
</script>

<a-entity networked="template: #avatar-template"></a-entity>
```

Results in:
```
<a-entity networked="template: #avatar-template">
  <a-entity gltf-model="#avatar-model">
    <a-text position="0 1 0" value="Avatar Name" ></a-text>
  </a-entity>
</a-entity>
```

This PR changes the following behaviors of templates:

1. Templates are defined in a `<template>` tag rather than `<script type="text/html">`
    This allows us to query and inspect the template before adding it to the DOM.

    Example:
    ```
    <template id="avatar-template">
      <a-entity gltf-model="#avatar-model">
        <a-text position="0 1 0" value="Avatar Name" ></a-text>
      </a-entity>
    </template>

    <a-entity networked="template: #avatar-template"></a-entity>
    ```

2. The default behavior of the `networked` component is to merge components of the root element of the template and append its children. This occurs both locally and remotely.

    The example above would result in:
    ```
    <a-entity networked="template: #avatar-template" gltf-model="#avatar-model">
      <a-text position="0 1 0" value="Avatar Name" ></a-text>
    </a-entity>
    ```

    This also means that the template *must only have one root element*. We can enforce this when defining the schema.

3. The `showLocalTemplate` property is removed.

    In order to hide the local template just use the `visible` component:
    ```
    <a-entity networked="template: #avatar-template" visible="false"></a-entity>
    ```

    The local template will be hidden and the remote will stay visible.

4. The `attachLocalTemplate` property is added.

    By setting `attachLocalTemplate: false` the template will not be attached to the local networked entity. This allows you to specify a different hierarchy for the local entity:
    ```
    <a-entity networked="template: #avatar-template; attachLocalTemplate: false;">
      <a-entity player-hud position="0 0 1"></a-entity>
    </a-entity>
    ```

    By default `attachLocalTemplate` is set to true.

5. `aframe-template-component` is removed as a dependency.

    `networked-aframe.min.js` shrinks from 67.8 kB to 37.8 kB.

### TODO:
- [ ] Validate templates when adding a schema
- [x] Update examples
- [x] Update tests


haydenjameslee:
Appreciate the nice description with examples ❤️

server/static/a-saturday-night/index.html
```diff
@@ -100,10 +100,22 @@
 
 			</a-assets>
 
-			<a-entity id="player" networked="template:#player-template;showLocalTemplate:false;" spawn-in-circle="radius:2;" wasd-controls show-avatar="index:0;numAvatars:3" randomize-avatar>
-			  <a-entity camera="userHeight: 1.6" look-controls networked="template:#head-template;showLocalTemplate:false;"></a-entity>
-			  <a-entity networked="template:#left-hand-template" hand-components="left" show-in-vr visible="false"></a-entity>
-			  <a-entity networked="template:#right-hand-template" hand-components="right" show-in-vr visible="false"></a-entity>
+			<a-entity id="player" networked="template:#player-template" spawn-in-circle="radius:2;" wasd-controls show-avatar="index:0;numAvatars:3" randomize-avatar>
+			  <a-entity camera="userHeight: 1.6" look-controls networked="template:#head-template"></a-entity>
+			  <a-entity networked="template:#left-hand-template" hand-components="left" show-in-vr visible="false">
```
haydenjameslee: Shouldn't these `networked` tags use `attachLocalTemplate: false;`? Because they are defining the child entities locally. In this case it seems like we can just remove the child entities however because they are the same as the template.

server/static/a-saturday-night/index.html
```diff
@@ -100,10 +100,22 @@
...
+          <a-entity class="heads" show-child="0">
+            <a-entity class="head" gltf-model="#avatar1-head-obj"></a-entity>
+            <a-entity class="head" gltf-model="#avatar2-head-obj"></a-entity>
+            <a-entity class="head" gltf-model="#avatar3-head-obj"></a-entity>
+          </a-entity>
+        </a-entity>
+			  <a-entity networked="template:#right-hand-template" hand-components="right" show-in-vr visible="false">
+          <a-entity class="heads" show-child="0">
+            <a-entity class="head" gltf-model="#avatar1-head-obj"></a-entity>
```
haydenjameslee: This is showing the head model instead of the hand model.

src/NetworkEntities.js
```diff
@@ -35,15 +35,15 @@ class NetworkEntities {
     var hasPosition = componentData.hasOwnProperty('position');
     if (hasPosition) {
       var position = componentData.position;
-      entity.setAttribute('position', position);
+      entity.setAttribute('position', `${position.x} ${position.y} ${position.z}`);
```
haydenjameslee: I thought A-Frame supported passing in objects rather than needing to serialize to a string here? Why the change?

robertlong: It does, but only after the `a-entity` Custom HTMLElement is done overwriting `setAttribute`.

I've created an example project to help myself and others learn about the Custom HTMLElement lifecycle.

https://aframe-custom-element-lifecycle.glitch.me/

Maybe we can make everything work asynchronously to deal with `setAttribute` not being overwritten until the next tick?

tests/unit/networked_attachLocalTemplate.test.js
```diff
@@ -0,0 +1,80 @@
...
+      assets: [
+        "<template id='t1'><a-entity><a-entity class='template-child'></a-entity></a-entity></template>"
+      ],
+      entity: '<a-entity id="test-entity" networked="template:#t1;attachLocalTemplate:false;" position="1 2 3" rotation="4 3 2"><a-box></a-box></a-entity>'
+    };
+    scene = helpers.sceneFactory(opts);
+    naf.utils.whenEntityLoaded(scene, done);
+  }
+
+  function MockNetworkAdapter() {
```
brianpeiris: This exists in `networked.test.js` and `NetworkConnection.test.js` as well. I would move it to `helpers.js` and reuse.

robertlong:
Should I create a new PR to merge this into `0.6.0-dev` or is `master` correct? There's a lot of noise in the diff right now from the `0.6.0-dev` branch.

brianpeiris:
You can see a cleaner diff here: https://github.com/networked-aframe/networked-aframe/compare/0.6.0-dev...netpro2k:feature/networked-templates-refactor

robertlong:
Looks like github allows you to change the base branch. That's cool.

src/NetworkEntities.js
```diff
@@ -18,43 +18,50 @@ class NetworkEntities {
...
 
   addNetworkComponent(entity, entityData) {
-    var networkData = {
-      template: entityData.template,
-      owner: entityData.owner,
-      networkId: entityData.networkId
-    };
-
-    entity.setAttribute('networked', networkData);
+    entity.setAttribute('networked', `template: ${entityData.template}; owner: ${entityData.owner}; networkId: ${entityData.networkId}`);
```
robertlong: I still really don't like this string serialization, but I'm unsure how to make everything work with the custom components lifecycle.

netpro2k:
Pretty happy with where we landed on this but it still bothers me how often we end up having to copy paste large parts of the template to the local side just to add some components. Maybe we will land somewhere where we do something similar to what we did  in our social client with `<a-gltf-entity>` and allow attaching stuff to things in the network template. But this is a great improvement and definitely simplifies some of the ugliness we had to do with physics in our social client.

src/components/networked.js
```diff
@@ -6,7 +6,7 @@ var bind = AFRAME.utils.bind;
 AFRAME.registerComponent('networked', {
   schema: {
     template: {default: ''},
-    showLocalTemplate: {default: true},
+    attachLocalTemplate: { default: true },
```
brianpeiris: This might be an annoying change at this point, so I'll leave it to your discretion. When I read `attachLocalTemplate`, it implies the concept of a "local template". Maybe this would be better named `attachTemplateLocally` or `attachTemplateToLocal` or `attachTemplateToLocalEntity`.

src/NetworkEntities.js
wmurphyrd: In A-Frame, entity and component initialization is synchronous now (as of 0.5.0 ish); you can use the object form of `setAttribute` immediately upon creation of an entity.

src/NetworkEntities.js
```diff
@@ -18,43 +18,50 @@ class NetworkEntities {
...
   initRotation(entity, componentData) {
     var hasRotation = componentData.hasOwnProperty('rotation');
     if (hasRotation) {
       var rotation = componentData.rotation;
-      entity.setAttribute('rotation', rotation);
+
+      if (typeof rotation === "string") {
+        entity.setAttribute('rotation', rotation);
+      } else {
+        entity.setAttribute('rotation', `${rotation.x} ${rotation.y} ${hasRotation.z}`);
```
wmurphyrd: I think you meant `rotation.z` not `hasRotation.z`. You could also safely pass rotation as an object

src/components/networked.js
```diff
@@ -93,45 +108,19 @@ AFRAME.registerComponent('networked', {
...
   },
 
   onConnected: function() {
     if (this.data.owner === '') {
       this.lastOwnerTime = NAF.connection.getServerTime();
       this.el.setAttribute(this.name, {owner: NAF.clientId});
-
-      this.waitForTemplate(() => {
+      setTimeout(() => {
+        //a-primitives attach their components on the next frame; wait for components to be attached before calling syncAll
```
robertlong: I've added an issue for this. We shouldn't need to do this.
https://github.com/aframevr/aframe/issues/3431

PR #94, opened 2018-02-08
Fix compatibility with the next version of a-frame

brianpeiris:
The next version of A-Frame changes the way camera pose updates are handled. A-Frame hands control over to three.js' WebVRManager in VR mode and does not set the camera entity's pose data like it did previously. Also, as a result of optimization, position/rotation/scale data is no longer updated in component data, so we should be using `getAttribute` to access that information.


server/static/basic.html
```diff
@@ -4,7 +4,7 @@
     <title>Dev Example — Networked-Aframe</title>
     <meta name="description" content="Dev Example — Networked-Aframe">
 
-    <script src="https://aframe.io/releases/0.7.0/aframe.min.js"></script>
+    <script src="https://cdn.rawgit.com/aframevr/aframe/c2c97c2a6dd470f2c24271bbb94529d12b7d3fb9/dist/aframe-master.js"></script>
```
brianpeiris: Pointed to aframe master temporarily for testing.

server/static/basic.html
```diff
@@ -92,7 +92,7 @@
         <!-- /Templates -->
       </a-assets>
 
-      <a-entity id="player" networked="template:#avatar-template;showLocalTemplate:false;" camera="userHeight: 1.6" spawn-in-circle="radius:3" wasd-controls look-controls>
+      <a-entity id="player" networked="template:#avatar-template;showLocalTemplate:false;" camera position="0 1.6 0" spawn-in-circle="radius:3" wasd-controls look-controls></a-entity>
```
brianpeiris: aframe vNext will not have a userHeight property on the camera. Instead, you need to set the position of the camera entity. 
Note that wasd-controls are also broken in VR mode, since WebVRManager has full control. This needs to be fixed by containing the camera in a camera/player rig. See https://github.com/aframevr/aframe/issues/3351

src/ComponentHelper.js
```diff
@@ -1,25 +1,40 @@
...
+        // However, the camera's world position incorporates the camera's rotation in a weird way, so we have to
+        // undo that rotation as well.
+        if (name === 'position' && el.components.hasOwnProperty('camera') && el.sceneEl.is('vr-mode')) {
+          el.components.camera.camera.getWorldPosition(cameraWorldPosition);
+          cameraGroupConjugate.copy(el.object3D.quaternion);
+          cameraGroupConjugate.conjugate();
+          cameraWorldPosition.sub(el.object3D.position);
+          cameraWorldPosition.applyQuaternion(cameraGroupConjugate);
+          cameraWorldPosition.add(el.object3D.position);
+          compsData[name] = AFRAME.utils.clone(cameraWorldPosition);
```
brianpeiris: This feels overly complicated. I'm going to find out if this is a mistake with the underlying code, but this works around things for now.

robertlong: Maybe we can add a method to the camera component in AFrame core to get its position vs doing it in libraries?

src/ComponentHelper.js
```diff
@@ -1,25 +1,40 @@
...
     } else {
       var childKey = NAF.utils.childSchemaToKey(element);
       var child = element.selector ? el.querySelector(element.selector) : el;
       if (child) {
-        var comp = child.components[element.component];
-        if (comp) {
-          var data = element.property ? comp.data[element.property] : comp.data;
+        if (child.components.hasOwnProperty(element.component)) {
+          var attributeData = child.getAttribute(element.component);
+          var data = element.property ? attributeData[element.property] : attributeData;
```
netpro2k: This can just become
```
var data = child.getAttribute(element.component, element.property);
```
since `getAttribute` can handle getting sub properties for you when an argument is passed.

src/ComponentHelper.js
netpro2k: Hmm yeah that fact that you would have to do this in a library feels very wrong. if the position/rotation can't be updated directly maybe this should be a special case handled in the position and rotation component's themselves.

brianpeiris: @dmarcos ^ for reference, this is what I have to do to get the camera's actual position now.

src/ComponentHelper.js
brianpeiris: Really? I don't see that documented anywhere.

haydenjameslee: I thought you could too but apparently not: https://github.com/aframevr/aframe/blob/master/src/core/a-entity.js#L726

I tried in the console too and it didn't work.

brianpeiris:
Good news! The camera transform is no longer required, since I've fixed the issue at the root cause (https://github.com/dmarcos/three.js/pull/2)

This PR is now also backwards compatible with the current version of A-Frame. So I think it's ready to merge.

src/ComponentHelper.js
netpro2k: hmm, could have sworn this used to work...

PR #111, opened 2018-03-30
support pos/rot/scale/vis not being components

ngokevin:
Quick fix for supporting pos/rot/scale/vis, my rotation wasn't syncing because it was checking component.data. pos/rot/scale/vis no longer are always components.

component.data is also an internal object and should use the getAttribute interface.

Still some perf stuff to fix later (no cloning, object allocations).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "networked-aframe",
-  "version": "0.5.1",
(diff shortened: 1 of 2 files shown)


brianpeiris:
This was fixed in #94 and is already merged into the 0.6 branch.

PR #114, opened 2018-04-01
Remove requirement to register all templates

haydenjameslee:
With the new template system in 0.6 we introduced a requirement to define a schema for every template that we use. In previous versions of NAF you didn't have to define a schema for a template if you just wanted `position` and `rotation` to be networked.

This PR re-introduces that ability by collecting all templates in the a-assets element and assigning them a default schema (which doesn't overwrite and can be overridden).


rajnikantmrao:
Hayden: can you add scale also as default that requires no schema?

haydenjameslee:
@rajnikantmrao Could you open that request as a separate issue? For the moment I just want to focus on solving the template system bug so I don't want to change any other existing functionality in the same branch.

PR #121, opened 2018-04-05
Use A-Frame's built-in deepEqual function instead of deep-equal module

haydenjameslee:


haydenjameslee:
Putting this on hold due to: https://github.com/MozillaReality/networked-aframe/pull/12

PR #122, opened 2018-04-06
Emit sound source and fix audio output for safari

brianpeiris:
This helps solve two issues with Safari:
1. If you want to use the audio stream in Safari, e.g to visualize the audio, you need access to the MediaStreamSource that networked-aframe creates, since Safari will play garbled audio if you try to `createMediaStreamSource` twice on the same stream. This PR adds an `sound-source-set` event that passes the sound source to any consumers.
2. Safari on iOS would play audio from the hidden `Audio` node that we were creating as a workaround for Chrome. This change applies the workaround to chrome browsers only.

### DIFF
--- a/src/components/networked-audio-source.js
+++ b/src/components/networked-audio-source.js
@@ -47,13 +47,19 @@ AFRAME.registerComponent('networked-audio-source', {
       }
       if(newStream) {
         // Chrome seems to require a MediaStream be attached to an AudioElement before AudioNodes work correctly
-        this.audioEl = new Audio();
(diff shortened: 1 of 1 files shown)


brianpeiris:
Note: I decided to use an event here, but @netpro2k proposed that we instead expose a public method on the `networked-audio-source` component that returns a promise which resolves with the sound source. Not sure what the best practice or preference is.

haydenjameslee:
Haven't had to think about the difference much yet, and I personally haven't had enough experience making actual apps to have a personal preference. Let's just go with this and see how it goes. Can always switch it later.

Issue #108, opened 2018-03-30
Handle position/rotation/scale/visible not being components.


ngokevin:
We removed position/rotation/scale/visible from being default components now that A-Frame can bypass .setAttribute for these.

networked-aframe should read directly from the object3D versus checking an entity's component set for these properties.

Issue #106, opened 2018-03-29
Don't sync pos/rot/scale via .setAttribute


ngokevin:
Use `el.object3D.position.set(x, y, z)` or whatever versus `el.setAttribute('position', data)`. 

The former has just about zero overhead. The latter has quite a bit. The data will still be synced on getAttribute. Save a lot of time on ticks.

haydenjameslee:
I think the only thing this would break is the existing `lerp` component, since it listens to an `oncomponentchanged` event which wouldn't get fired in this optimization. That being said however, we're looking into changing how that lerp component works to make it more explicit. We could make both changes at the same time.

ngokevin:
I wouldn't use `componentchanged` for anything, particularly for rapidly changing stuff. pos/rot will always be changing, I'd run a continuous tick. Yeah, lerp should stop using setAttribute (and most ticks for that matter) and go for the object3D API.

haydenjameslee:
Oh i was wrong, turns out the lerp doesn't use the component changed event anymore: https://github.com/haydenjameslee/aframe-lerp-component/blob/master/index.js#L41

ngokevin:
Right, I think we've talked about that before haha. That could have some configurable check to not do the `checkForComponentChanged` in some cases...like head pose or controller pose is always going to change.

Okay, so on the NAF side, this isn't a problem, because it always uses lerp for pos/rot?

haydenjameslee:
Its still a problem, it basically changes the position _twice_ because of this. Once via .setAttribute and once via object3D. NAF uses .setAttribute then the lerp comp uses object3D. Instead we are going to remote that lerp component and write our own NAF lerper that uses object3D directly.

haydenjameslee:
@netpro2k do you have a prototype of this new lerper in action already?

netpro2k:
Nothing beyond what we discussed in the past on slack, still not sure what the best path forward with that is.

haydenjameslee:
+1

vincentfretin:
In current code, when lerping is used, tick uses `object3D.position.copy(position)`, so I'm closing this issue.

PR #124, opened 2018-04-10
Moved lerping into the networked component.

robertlong:
Addresses: #106 and https://github.com/haydenjameslee/aframe-lerp-component/issues/3

This PR should also probably address: #108 

TODO:
- [x] Add NAF.options.useLerp back in
- [x] Clean up code
- [ ] Fix tests


src/components/networked.js
```diff
@@ -302,16 +323,89 @@ AFRAME.registerComponent('networked', {
...
+          this.updateComponent(el, key, data);
         }
       }
     }
   },
 
+  updateComponent: function (el, key, data) {
+    switch(key) {
+      case "position":
+        var posComp = this.positionComponents.find((item) => item.el === el);
```
netpro2k: I could see this find becoming problematic with ltos of components so we may want to index at some point but its probably fine for now.

src/options.js
```diff
@@ -2,6 +2,5 @@ var options = {
   debug: false,
   updateRate: 15, // How often network components call `sync`
   compressSyncPackets: false, // compress network component sync packet json
-  useLerp: true, // when networked entities are created the aframe-lerp-component is attached to the root
```
netpro2k: Getting rid of this flag is probably reasonable but there obviously needs to be a way to control this at the schema level. We should think about how that configuration will be extended later to support other interpolation models for other components.

src/components/networked.js
```diff
@@ -302,16 +323,89 @@ AFRAME.registerComponent('networked', {
...
+        if (!posComp) {
+          posComp = {};
+          posComp.el = el;
+          posComp.start = new THREE.Vector3(data.x, data.y, data.z);
+          posComp.target = new THREE.Vector3(data.x, data.y, data.z);
+          posComp.lastUpdated = Date.now();
+          posComp.duration = 1;
+          this.positionComponents.push(posComp);
+        } else {
+          posComp.start.copy(posComp.target);
```
netpro2k: This is the same lerping strategy that `aframe-lerp-component` uses but is only valid for data that is continously being updated. For objects that aren't continously moving this results in some pretty undesirable behavior. Since this isn't a regression it's fine for this PR but we definitely need to think of ways to allow other interpolation strategies.

src/components/networked.js
```diff
@@ -162,6 +156,34 @@ AFRAME.registerComponent('networked', {
...
       this.syncDirty();
     }
+
+    var now = Date.now();
+
+    if (!this.isMine()) {
+      for (const posComp of this.positionComponents) {
+        var progress = (now - posComp.lastUpdated) / posComp.duration;
+        
+        if (progress <= 1) {
```
netpro2k: Looks like this actually wont ever fire on the initial update since progress ends up being > 1 (from the duration being 1ms)

haydenjameslee:
Yeah the tests aren't very good, I have similar effects happen quite often. I'm not sure what the root issue is.

src/components/networked.js
```diff
@@ -319,18 +320,32 @@ AFRAME.registerComponent('networked', {
               childEl.setAttribute(schema.component, schema.property, data);
             }
             else {
-              this.updateComponent(childEl, schema.component, data);
+              var shouldLerp = this.shouldLerpComponent(syncedComponents, schema);
```
netpro2k: I am confused, can't you just check `schema.lerp` here?

src/components/networked.js
```diff
@@ -319,18 +320,32 @@ AFRAME.registerComponent('networked', {
...
+      if (schema.selector === syncedComponent.selector && schema.component === syncedComponent.component) {
+        return syncedComponent.lerp !== false;
+      }
+    }
+
+    return false;
+  },
+
+  updateComponent: function (el, key, data, lerp) {
+    if (!NAF.options.useLerp || !lerp) {
```
netpro2k: I'd remove this check here and instead do it in each case so that you can still use the direct setter:
```
if(!useLerp) {
    el.object3D.position.set(data.x, data.y, data.z);
} else if(!posComp) {
    ...
} else {
    ...
}
```

src/components/networked.js
robertlong: The return value of `NAF.utils.keyToChildSchema` isn't actually a schema. So you cant pull `schema.lerp` off of it. I'd like to rename schema to something else in this context but I'm not really sure what to call it.

InfiniteLee:
@robertlong Can we close this in favor of #130?

PR #127, opened 2018-05-07
Change package.json to expose index.js

robertlong:
networked-aframe currently exposes the bundled/dist version of networked-aframe in package.json. Including the bundled version in a webpack project makes it so the source map generated points to the sourcemap of the bundle and not the individual files of the library. The downsides of doing this is that all consumers of the library need to have ES2015 support. I think it's safe now to require ES2015 as most developers are using Babel if they are requiring the npm package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.6.1",
   "description": "A web framework for building multi-user virtual reality experiences.",
   "homepage": "",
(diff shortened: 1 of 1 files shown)


haydenjameslee:
Yup makes sense. I also think most people are referencing the dist via unpkg.com anyway (or at least that's what I encourage).

PR #141, opened 2018-07-24
Bug/accidental interpolation buffers

johnshaughnessy:
To see the bug, run the example page on mr-social-client/master before pulling in this change.

The bug manifests itself in a couple ways:
- If your schema says you'll sync `position`, `rotation`, or `scale`, then the `networked` component will read ALL of them from the interpolation buffer. 
- If your template has a child with any synced components, the `position`, `rotation`, and `scale` will be set to the default values every tick for the same reason. In the example page, this manifests itself in the following way:
The `.head` entity's `scale` will initialize to `{x: 0.45, y: 0.5, z: 0.5}` as per its attribute values declared in the `#avatar-template`.
The `networked` component will create an `InterpolationBuffer` for the `.head` entity in its first `networkUpdate` (via `updateComponents` -> `updateComponent`), even though the `.head` entity only wants to sync its `material.color`. 
On the first `tick` of the `networked` component, the `position`, `rotation`, and `scale` of the `interpolationBuffer` created for the `.head` entity will overwrite the value in the `head`'s `object3D`.
(Note that at this point, `headEntity.components.scale.data === {x: 0.45, y:0.5, z:0.5}` but `headEntity.object3D.scale === {x: 1, y: 1, z:1}`. It is scary that this is possible, but I suppose we are willing to take this risk to avoid calling `setAttribute`.)

This patch tries to fix the issues above.

See https://github.com/MozillaReality/networked-aframe/pull/15 for discussion / history.

(Sorry to point to a PR in the MozillaReality repo. I should have made the PR against this repo's master branch. I'll do that in the future.)

### DIFF
--- a/src/components/networked.js
+++ b/src/components/networked.js
@@ -393,12 +393,13 @@ AFRAME.registerComponent('networked', {
                      object3D: el.object3D,
                      componentNames: [componentName] };
       this.bufferInfos.push(bufferInfo);
+    } else {
+      var componentNames = bufferInfo.componentNames;
+      if (!componentNames.includes(componentName)){
+        componentNames.push(componentName);
+      }
     }
     var buffer = bufferInfo.buffer;
-    var componentNames = bufferInfo.componentNames;
-    if (!componentNames.includes(componentName)){
-      componentNames.push(componentName);
-    }
 
     switch(componentName) {
       case "position":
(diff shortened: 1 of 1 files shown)


haydenjameslee:
Oh whoops I thought I merged this the other night. Merging now.

PR #146, opened 2018-08-15
Correct networked entity deletion event name

3DataWill:
The event type that is fired is `'entityRemoved'`

https://github.com/networked-aframe/networked-aframe/blob/4c270ff0dc5b6429293c110141697fa326fdef0b/src/components/networked.js#L441


disregardfiat:
come together right now

PR #149, opened 2018-09-15
Fix issues with custom delta comparators

gfodor:
This PR changes the contract for `requiresNetworkUpdate` custom comparator predicates to be pure comparator functions. Previously, it was hard to write a stateful comparator since, unlike the default, there was no way to create new bindings to keep track of previous state. This PR now keeps track of the previous sync values in the networked component directly and passes it to the comparator to check if it needs to be updated.

### DIFF
--- a/src/components/networked.js
+++ b/src/components/networked.js
@@ -4,19 +4,6 @@ var InterpolationBuffer = require('buffered-interpolation');
 var DEG2RAD = THREE.Math.DEG2RAD;
 var OBJECT3D_COMPONENTS = ['position', 'rotation', 'scale'];
 
-function defaultRequiresUpdate() {
-  let cachedData = null;
-
-  return (newData) => {
-    if (cachedData === null || !deepEqual(cachedData, newData)) {
-      cachedData = AFRAME.utils.clone(newData);
-      return true;
-    }
-
-    return false;
-  };
-}
-
 AFRAME.registerComponent('networked', {
   schema: {
     template: {default: ''},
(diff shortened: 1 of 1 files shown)


src/components/networked.js
```diff
@@ -287,8 +275,14 @@ AFRAME.registerComponent('networked', {
...
 
       // Use networkUpdatePredicate to check if the component needs to be updated.
-      // Call networkUpdatePredicate first so that it can update any cached values in the event of a fullSync.
-      if (this.networkUpdatePredicates[i](syncedComponentData) || fullSync) {
+      const previousData = this.previousSyncedComponentData[i];
+      const hasChanged = previousData === null || this.networkUpdatePredicates[i](previousData, syncedComponentData);
+
+      if (hasChanged) {
+        this.previousSyncedComponentData[i] = AFRAME.utils.clone(syncedComponentData);
+      }
```
mquander: I think the point of this whole exercise was to track prior data on a type-by-type basis, so that we can do something more efficient than cloning every time (e.g. for vectors we want an implementation where each vector property has a single prior `Vector3` that it reuses every network frame.)

mquander:
Nice find -- John and/or Robert will think this is funny when they see it, because I'm 80% sure this bug wasn't in the original version and then I snuck it in there when I was bringing this branch up-to-date prior to merging. I remember some conversation where they were like "but you changed the contract of the network predicates" and I was like "oh sure, look how easy mine is" and they were like "wow, how wonderful is what you did!"

I don't think your fix is quite what we want, because of the cloning? I liked your idea from Slack where we say that `requiresNetworkUpdate` returns a per-component function that closes over any state, instead of being itself a function that closes over state (which as you observed is wrong since there's just one per schema.) I think that's relatively straightforward.

gfodor:
Haha cool, that is fine with me. I will write up an alternative PR

gfodor:
https://github.com/networked-aframe/networked-aframe/pull/150

PR #150, opened 2018-09-17
Assume requiresNetworkUpdate is a second order function

gfodor:
This PR changes the contract for requiresNetworkUpdate custom predicate on the schema to now be a second order function, like the default, so it can create component instance-level bindings for state (like as is done in the default for the previous value.)


mquander:
OK, this one I endorse completely :+1:

gfodor:
Closed for https://github.com/networked-aframe/networked-aframe/pull/151

PR #152, opened 2018-09-19
A simple take on networked entity persistence

3DataWill:
Sometimes you don't want an entity to be destroyed when its owner disconnects. This could be a model or drawing added to a scene by a user or a singleton object like a presentation controller.

This approach adds the `persistent` property to the `networked` component, and, on remote owner disconnect, attempts to take ownership of persistent entities rather than delete them. It's not the most efficient approach, as, in a room with multiple users, everyone races for ownership and the slowest wins. It is, however, achieved with minimal changes and should work even if the original owner crashes or otherwise disconnects unexpectedly.

If you're interested in this approach, I can add documentation and testing to this PR. If you're not, no sweat, just wanted to share what we're using.

### DIFF
--- a/src/NetworkEntities.js
+++ b/src/NetworkEntities.js
@@ -50,7 +50,8 @@ class NetworkEntities {
     var networkData = {
       template: entityData.template,
       owner: entityData.owner,
-      networkId: entityData.networkId
+      networkId: entityData.networkId,
+      persistent: entityData.persistent
     };
 
     entity.setAttribute('networked', networkData);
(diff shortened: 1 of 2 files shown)


gfodor:
Thank you for writing this! I was about to sit down and implement a similar solution and was amazed to see you had already done it :)

gfodor:
This patch should fix up the tests. I'm having trouble getting the full suite to pass on Travis CI atm (works locally) but this fixes the main unit tests.

gfodor:
https://github.com/MozillaReality/networked-aframe/pull/18/commits/cfcd671833aa1617cd7049e98685f0635051eae1

gfodor:
one more commit to get tests to pass: https://github.com/MozillaReality/networked-aframe/pull/18/commits/ef635f410efee1fcf7eb8230e691d91b80d35e48

adding a proper test case proved rather difficult, since the `NetworkedEntities` test suite seems to not actually bootstrap up a proper component. might revisit this later.

Issue #101, opened 2018-03-08
easyrtc: enable audio receive even if no getMediaStream supported


antoniohof:
I suggest setting `this.easyrtc.enableAudioReceive(true);` as peers should always receive audio, regardless of them having access to mic or a supported browser.

This change includes setting the 	      `this.easyrtc.setStreamAcceptor(this._storeAudioStream.bind(this));
` 
outside the _connectWithAudio function, as it must be set even if you didnt connect with audio: true.

antoniohof:
I will make a pull request on the weekend.

antoniohof:
I'm working on a pull request

haydenjameslee:
Awesome, that make sense.

antoniohof:
https://github.com/networked-aframe/networked-aframe/pull/159

haydenjameslee:
Merged. I appreciate the PR!

PR #159, opened 2019-06-11
easyrtc adapter - accept incoming audio stream even with audio disabled

antoniohof:
modified easyrtc adapter to accept incoming audio streams even if you don't have audio enabled.


antoniohof:
fixes https://github.com/networked-aframe/networked-aframe/issues/101

PR #182, opened 2020-03-30
Fix getServerTime returning a string

zach-capalbo:
On Chrome and Firefox, `new Date() + this.avgTimeOffset` results in a date string with `avgTimeOffset` appended as a string to the end.

This pull request converts the date to a number, which ensures that addition happens numerically, ensuring that `getServerTime` returns a time number, as expected (as evidenced by comparing `getServerTime()` using less than and greater than elsewhere in the code)

### DIFF
--- a/src/adapters/naf-socketio-adapter.js
+++ b/src/adapters/naf-socketio-adapter.js
@@ -216,7 +216,7 @@ class SocketioAdapter {
   }
 
   getServerTime() {
-    return new Date() + this.avgTimeOffset;
+    return new Date().getTime() + this.avgTimeOffset;
   }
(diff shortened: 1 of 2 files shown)


andronat:
Hello @zach-capalbo could you also please fix the same issue [here](https://github.com/networked-aframe/networked-aframe/blob/master/src/adapters/naf-webrtc-adapter.js#L585) please? It's for the WebRTC adapter.

zach-capalbo:
@andronat, added a change to fix the issue on the WebRTC adapter

haydenjameslee:
Thanks for this!

Issue #200, opened 2020-05-15
Avatars disappear in demos


osresearch:
I've setup the demos in a clean checkout with a nginx reverse proxy doing SSL termination and it almost works perfectly. In the basic and tracked controller demo, however, users (including VR headsets) can connect to the room and see the other players, but after a little while some of the other avatars disappear from view (but not from the occupants array).  It's often very disjoint as to who can still see who. Reloading the page fixes things for a little while. 

There is an error on the console about `this.data` being undefined in `networkUpdate()`, but it appears before the users start disappearing. There don't appear to be any console messages when the avatars go missing.

Any ideas for what might be happening?

osresearch:
This happens with both `adapter: socketio` and `adapter: webrtc`.

mneunomne:
I have exactly the same error in my application. I am currently using socketio adapter.
After the error the client keeps disconnecting from the server with the reason being "ping timeout".

mneunomne:
I was able to consistently reproduce this error by connecting a mobile device and be constantly moving the gyroscope so there is always an updating happening in the scene. While i do that and try to connect another client, "this.data" is undefined from the update coming from that mobile client.

I was able to make avoid this exception by simply returning the function if `data` is undefined with the following code on line 409 in the `src/components/networked.js` file:

` if (this.data === undefined) {
      return
    }`

Apparently `this.data` is undefined for a couple of times while this function is called, then the component seems to be correctly loaded. Maybe it is an issue of receiving updates from a component that was not yet locally loaded ?

Surely there is a better solution than this little hack i made, but it's temporarily working for me  😅.

Cheers!

osresearch:
The patch in #202 might not have fixed it, unfortunately.  With the fix applied and using webrtc, sometimes avatars disappear, although it seems very asymmetric since they can still see the other players. On the console of the computer that no longer sees the other there is a message:

```
Closed data channel from goE7AIlw7E6ZeiG5AAAF
```

vincentfretin:
Still with socketio? This seems that the socket timed out and closed, no? You may need to send regularly something in the socket so it stays alive.

vincentfretin:
[...]

PR #202, opened 2020-05-16
components/networked.js: this.data can be undefined

osresearch:
Sometimes the `this.data` value is undefined if there is a high load of messages coming in, such as when a VR headset joins. This causes the other avatar to disappear from the screen, even though it still appears in the occupants array and the other player can still see this one.

Discussion in issue #200

### DIFF
--- a/src/components/networked.js
+++ b/src/components/networked.js
@@ -405,6 +405,7 @@ AFRAME.registerComponent('networked', {
       return;
     }
 
+    // Hack to solve this bug: https://github.com/networked-aframe/networked-aframe/issues/200
     if (this.data === undefined) {
       return;
     }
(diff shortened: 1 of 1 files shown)


src/components/networked.js
```diff
@@ -405,6 +405,10 @@ AFRAME.registerComponent('networked', {
       return;
     }
 
+    if (this.data === undefined) {
```
haydenjameslee, line 408:
```suggestion
    // Hack to solve this bug: https://github.com/networked-aframe/networked-aframe/issues/200
    if (this.data === undefined) {
```

haydenjameslee:
I'll merge it for now and we can solve the root cause at a later date

vincentfretin:
I'll just mention it here. Mozilla did a similar fix in `isMine` https://github.com/MozillaReality/networked-aframe/pull/44
I'll retrieve this change as part of #241 too.

Issue #194, opened 2020-05-10
[Bug]  One-way audio transmission with WebRTC adapter


djoksimo:
Hi 👋 ,

I tried out https://haydenlee.io/networked-aframe/ and I also ran the [example](https://github.com/networked-aframe/networked-aframe/blob/master/examples/basic-audio.html
) from source:
`git clone, npm install, npm run dev` (served over HTTPS)

I tested this [example that uses the EasyRTC adapter instead of the WebRTC adapter](https://glitch.com/edit/#!/easyrtc-version) and it seems to work (but not on iOS or Safari) and it's deprecated (as mentioned in your repo).

## Steps
1. Set up basic socket.io signalling as shown in `server.js` ([file link](https://github.com/networked-aframe/networked-aframe/blob/master/server/index.js) in the repo)
2. Set up `networked-scene` attributes with `adapter: webrtc` and `audio: true`
3. Ran https local server and opened app with same roomId on two different tabs
4. The first tab was the only recognized "audio source" and was the only one transmitting audio
5. Second tab can hear the first one, but is not able to transmit audio

**TL:DR; CANNOT HEAR ECHO**

### Notes
1. The avatars appear
2. we can track each other's movement/rotation
3. only one person can be heard (the first person to join the room)

silverslade:
H!
I confirm this issue.
Temporary switched to the 0.6.0 version with open easyrtc to get things work.

haydenjameslee:
I'll have to look into this. Thanks for bringing it to my attention. PRs welcome.

osresearch:
I'm encountering the same issue with patch #202 applied to keep all of the players visible.

silverslade:
Hi !
I've done some further tests.
It occurs on the haydenlee.io page also: the first which enters in the room can listen and speak with microphone, the second which enters in the room can only listen ( ref: https://haydenlee.io/networked-aframe/basic-audio.html )
Tested on Firefox 76.0.1 and Chrome 81.0.4044.138 on Windows 10. 
Same issue: only the first user can listen and speak.
No problems with open easyrtc and networked-aframe 0.6.0 (the earlier version) with adapter: webrtc and audio: true

djoksimo:
It seems as though `that.pendingAudioRequest[clientId]` never gets resolved when the second user joins:

https://github.com/networked-aframe/networked-aframe/blob/130687dc49859c8baf8de61176fbbb4249405ee9/src/adapters/naf-webrtc-adapter.js#L555

PR #206, opened 2020-05-24
Fix 1-way audio bug

djoksimo:
Fixes the 1-way audio issue:

TL;DR: only one user can talk with the native webrtc adapter. This PR should fix this bug.

Closes https://github.com/networked-aframe/networked-aframe/issues/194

Adding this snippet seems to do the trick:

```javascript
             localStream.getTracks().forEach(
                track => {
                  Object.keys(self.peers).forEach(peerId => { 
                  self.peers[peerId].pc.addTrack(track, localStream) 
                })
              })
```

Safari doesn't seem to work (before & after this PR), but the [commented link](https://github.com/OpenVidu/openvidu/blob/master/openvidu-browser/src/OpenViduInternal/WebRtcPeer/WebRtcPeer.ts#L154) in the webrtc adapter file seems useful. I'll try to fix Safari functionality in this PR as well.

Tested by:

- opening two tabs in Chrome for localhost:8080/basic-audio.html.
- served over HTTPS using ngrok
- heard echo with earbuds

In 1-tab:

Already had audio for Is3UNSNjGAK2Jr4pAAAE in 1st tab.
Already had audio for E3Yb914FUJgxNX5pAAAD in 2nd tab.

I think that we can prevent duplicating redundant function calls:

e.g.:
> Already had ...

But this would require some exploration

### DIFF
--- a/src/adapters/naf-webrtc-adapter.js
+++ b/src/adapters/naf-webrtc-adapter.js
@@ -370,6 +370,7 @@ class WebrtcAdapter {
           .then(localStream => {
             self.storeAudioStream(self.myId, localStream);
             self.connectSuccess(self.myId);
+                  self.peers[peerId].pc.addTrack(track, localStream) 
           })
           .catch(e => {
             NAF.log.error(e);
--- a/src/adapters/naf-webrtc-adapter.js
+++ b/src/adapters/naf-webrtc-adapter.js
@@ -20,7 +20,7 @@ class WebRtcPeer {
   }
 
   offer(options) {
-    var self = this;
+    const self = this;
     // reliable: false - UDP
     this.setupChannel(
       this.pc.createDataChannel(this.channelLabel, { reliable: false })
@@ -103,8 +103,8 @@ class WebRtcPeer {
    */
 
   createPeerConnection() {
-    var self = this;
-    var RTCPeerConnection =
+    const self = this;
+    const RTCPeerConnection =
       window.RTCPeerConnection ||
       window.webkitRTCPeerConnection ||
       window.mozRTCPeerConnection ||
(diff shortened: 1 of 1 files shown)


vincentfretin:
Not related to native webrtc adapter per se, I use on my project mozilla janus adapter and was having issue not hearing audio on Safari, I applied the hack that Hubs is doing for this issue, you can find the comment and code here: https://github.com/mozilla/hubs/blob/0cf532c509ca792438952414968cbe814f21b1b5/src/hub.js#L1426-L1470

src/adapters/naf-webrtc-adapter.js
```diff
@@ -367,16 +367,22 @@ class WebrtcAdapter {
...
-            self.storeAudioStream(self.myId, localStream);
-            self.connectSuccess(self.myId);
-          })
-          .catch(e => {
-            NAF.log.error(e);
-            console.error("Microphone is disabled due to lack of permissions");
-            self.sendAudio = false;
-            self.connectSuccess(self.myId);
-          });
+            .then(localStream => {
```
osresearch, line 370: Is the white space change a spurious diff?

osresearch:
With this patch applied this seems to work! Thanks for tracking down the issue.

Although testing it took me down a debugging path -- the webrtc demo doesn't request mic permissions; so none of the participants had any audio.

src/adapters/naf-webrtc-adapter.js
djoksimo: I added this line as well:

```
 localStream.getTracks().forEach(
                track => {
                  Object.keys(self.peers).forEach(peerId => { 
                  self.peers[peerId].pc.addTrack(track, localStream) 
                })
              })
```

I'll revert the whitespace diff

PR #229, opened 2020-10-09
Fix link to tracked controller example

andybak:


vincentfretin:
Thank you @andybak
Merging it!

PR #234, opened 2021-01-19
Fixed socket adaptor error when disconnect method called

### LINKED ISSUE #215: cannot disconnect by using method in readme
AFRAME.scenes[0].removeAttribute('networked-scene');
```
networked-scene disconnected
Uncaught TypeError: this.adapter.disconnect is not a function
    at NetworkConnection.disconnect (NetworkConnection.js:248)
    at NewComponent.remove (networked-scene.js:75)
    at HTMLElement.removeComponent (a-entity.js:399)
    at HTMLElement.removeAttribute (a-entity.js:513)
    at <anonymous>:1:18
disconnect	@	NetworkConnection.js:248
remove	@	networked-scene.js:75
removeComponent	@	a-entity.js:399
removeAttribute	@	a-entity.js:513
(anonymous)	@	VM3870:1
```

suresh-borad:
**PR Changes:**
Used Method: `NAF.connection.disconnect()`
Error: `TypeError: this.adapter.disconnect is not a function`

Fixed error #215 

closes #215

### DIFF
--- a/src/adapters/naf-socketio-adapter.js
+++ b/src/adapters/naf-socketio-adapter.js
@@ -66,43 +66,43 @@ class SocketioAdapter {
           self.wsUrl = "ws://" + location.host;
         }
       }
-  
+
       NAF.log.write("Attempting to connect to socket.io");
       const socket = self.socket = io(self.wsUrl);
-  
+
       socket.on("connect", () => {
         NAF.log.write("User connected", socket.id);
         self.myId = socket.id;
         self.joinRoom();
       });
-  
+
       socket.on("connectSuccess", (data) => {
         const { joinedTime } = data;
-  
+
         self.myRoomJoinTime = joinedTime;
         NAF.log.write("Successfully joined room", self.room, "at server time", joinedTime);
 
         self.connectSuccess(self.myId);
       });
-  
+
       socket.on("error", err => {
         console.error("Socket connection failure", err);
         self.connectFailure();
       });
-  
+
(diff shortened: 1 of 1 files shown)


vincentfretin:
Thanks @suresh-borad for the fix. I'm trying to take over the maintenance of this repository now that @haydenjameslee gave me the permission to do so.
I'm not used to socketio.
@kylebakerio proposed `NAF.connection.adapter.socket.io.disconnect()` is there a difference?
What about `NAF.connection.adapter.socket.close()`?

kylebakerio:
The docs mention a generalized method for disconnecting no matter what adapter you use.

Using the socket.close() method only works when using the socket adapter.

The other option is removing the networked-scene component  from the scene, iirc. (Typing from phone and memory right now.)

suresh-borad:
@kylebakerio 
Yes you are right,
I have fixed the error for socket adapter, that i have used in my project.

vincentfretin:
Yes I understand the `disconnect()` function needs to be implemented in each adapter. My question was if `socket.disconnect()` is the right thing to do for the socketio adapter. Looking at the socket.io-client code, it appears that yes, there exist a `close()` function as well but it's just an alias for `disconnect()`, see https://github.com/socketio/socket.io-client/blob/47f917afdd1821079723542f081c726596f2aaf3/lib/socket.ts#L420-L444
I was just asking because I'm only familiar with the low level `WebSocket.close()`.
So ok your change is good, I merge. Thank you!

Issue #169, opened 2019-09-18
serve https


dirkk0:
Hi,

  first of all: thanks for this great library!

To test with multiple devices, it would be better to serve the files via https, because audio otherwise will silently (sic!) fail due to security reasons.

The only thing one needs to do is to insert:

    const https = require("https");
    const fs = require("fs");
    var privateKey  = fs.readFileSync('key.pem', 'utf8');
    var certificate = fs.readFileSync('cert.pem', 'utf8');
    var credentials = {key: privateKey, cert: certificate};

add some self created keys, change the http references to https and start the server with the credentials:

    var webServer = https.createServer(credentials, app);

I would create a pull request but I am not sure if we should change the existing easyrtc-server.js or better create a new one.

Best,
  Dirk

TheBricktop:
easy-rtc is already in dire need of being replaced because of multiple security flaws, so go ahead and create pull.

hthetiot:
See Open-EasyRTC SSL sample:
- https://github.com/open-easyrtc/open-easyrtc/blob/master/server_example/server_ssl.js#L1

Self Signed certificates for testing:
- https://github.com/open-easyrtc/open-easyrtc/tree/master/server_example/certs

sikaar:
Hi, 
I am not a developper but got this to work properly, serving https and redirecting http properly.  by updating /server/index to the following (remember to update the path to for var privateKey & var certificate)

`// Load required modules
const http = require("http"); // http server core module
const https = require("https");
const fs = require("fs");
const path = require("path");
const express = require("express"); // web framework external module


var privateKey  = fs.readFileSync('../key.pem', 'utf8');
var certificate = fs.readFileSync('../cert.pem', 'utf8');
var credentials = {key: privateKey, cert: certificate};


// Set process name
process.title = "networked-aframe-server";

// Get port for http or default to 80 and redirect to https
const httpApp = express();
httpApp.set('port', process.env.PORT || 80);
httpApp.get("*", function (req, res, next) {
    res.redirect("https://" + req.headers.host + "/" + req.path);
});

// Setup and configure Express https server.
const app = express();
app.set('port', process.env.PORT || 443);

app.use(express.static(path.resolve(__dirname, "..", "examples")));

// Serve the example and build the bundle in development.
if (process.env.NODE_ENV === "development") {
  const webpackMiddleware = require("webpack-dev-middleware");
  const webpack = require("webpack");
  const config = require("../webpack.dev");

  app.use(
[...]

Issue #244, opened 2021-02-27
Add a nginx conf example in the documentation


vincentfretin:
Add in the documentation a nginx conf example to deploy the nodejs server behind nginx with tls for production.
see https://github.com/open-easyrtc/open-easyrtc/issues/61#issuecomment-786649791
PR wanted :)

Note that you can also configure nodejs with a certificate, see #169

vincentfretin:
Example of a nginx config I use:
```nginx
server {
  listen      80;
  listen      [::]:80;
  server_name www.mydomain.com;
  return 301 https://mydomain.com$request_uri;
}

server {
  listen      443 ssl http2;
  listen      [::]:443 ssl http2;
  server_name www.mydomain.com;

  # https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=modern&openssl=1.1.1k&guideline=5.6
  ssl_certificate /etc/letsencrypt/live/mydomain.com/fullchain.pem;
  ssl_certificate_key /etc/letsencrypt/live/mydomain.com/privkey.pem;
  ssl_session_timeout 1d;
  ssl_session_cache shared:MozSSL:10m;  # about 40000 sessions
  ssl_session_tickets off;

  # modern configuration
  ssl_protocols TLSv1.3;
  ssl_prefer_server_ciphers off;

  # HSTS (ngx_http_headers_module is required) (63072000 seconds)
  add_header Strict-Transport-Security "max-age=63072000" always;

  # OCSP stapling
  ssl_stapling on;
  ssl_stapling_verify on;

  # verify chain of trust of OCSP response using Root CA and Intermediate certs
  ssl_trusted_certificate /etc/letsencrypt/live/mydomain.com/chain.pem;
  resolver 8.8.8.8 8.8.4.4;

  return 301 https://mydomain.com$request_uri;
}

server {
  listen      80 default_server;
  listen      [::]:80 default_server;
  server_name mydomain.com;
  # allow letsencrypt
  location ~ /\.well-known {
    allow all;
    root /var/www/html;
    try_files $uri $uri/ =404;
  }
  location / {
    return 301 https://$host$request_uri;
  }
}

server {
  listen      443 ssl http2;
  listen      [::]:443 ssl http2;
  server_name mydomain.com;
  keepalive_timeout   70;
  # allow letsencrypt
  location ^~ /\.well-known {
    allow all;
    root /var/www/html;
    try_files $uri $uri/ =404;
  }
  location /socket.io {
[...]

PR #257, opened 2021-02-28
Update hosting-networked-aframe-on-a-server.md

dirkk0:
added code

refs #169

### DIFF
--- a/docs/hosting-networked-aframe-on-a-server.md
+++ b/docs/hosting-networked-aframe-on-a-server.md
@@ -66,6 +66,19 @@ To integrate XirSys into your hosted NAF instance follow these steps:
 
 If your experience uses WebRTC audio functionality, you need to setup an SSL certificate for your domain name (so your domain uses https instead of http). This is neccessary because web browsers enforce a strict rule that in order to use WebRTC audio streaming your web page must be encrypted. This is to stop man-in-the-middle attacks from being able to take control of your computer's microphone or webcam.
 
+To acchieve this one needs to insert:
+
+    const https = require("https");
+    const fs = require("fs");
+    var privateKey  = fs.readFileSync('key.pem', 'utf8');
+    var certificate = fs.readFileSync('cert.pem', 'utf8');
+    var credentials = {key: privateKey, cert: certificate};
+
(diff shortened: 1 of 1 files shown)


vincentfretin:
Hi, sorry I'm not merging this right now. I don't think the instructions are enough for a complete beginner, you don't explain how to generate the certificate.

I'm curious, are you using tls termination on nodejs for production, exposing the nodejs port directly?
I know someone that do the tls termination on nodejs but expose nodejs behind IIS on Windows server.
I usually put nodejs behind nginx or traefik and do the tls termination on nginx or traefik.

In the getting started tutorial, there is "Using HTTPS locally:" https://github.com/networked-aframe/networked-aframe/blob/master/docs/getting-started-local.md#voice-chat--audio-streaming where we explain to use ngrok.
I think a good alternative is using your above code with a certificate generated with [mkcert](https://github.com/FiloSottile/mkcert) so to explain all that in the getting started tutorial.

For production deployment so in the https://github.com/networked-aframe/networked-aframe/blob/master/docs/hosting-networked-aframe-on-a-server.md document , we can reference those instructions again, but we should probably explain how to configure nginx #244

dirkk0:
Hi Vincent, np at all!

I usually use Letsencrypt to create a real certificate, since I mainly use this on either an AWS instance or a Raspi for temporary installations. I don't have a production environment/use case atm, but I would also be interested on how to so a proper setup with nginx and the likes.

For an experimental environement using Letsencrypt seems easy enough, and I could document this, if you feel this is useful.

I can't comment much on mkcert but if this doesn't create a 'real' certificate it is useless for my scenarios.

vincentfretin:
For a deployment with nginx, I usually use certbot to generate a letsencrypt certificate so yeah it would be good to document it.
For mkcert, it's not a real certificate for production, it's only for development. It's better than a self generated certificate generated with openssl command for example because it adds a CA to your system and create the certificate with this registered CA, so you don't have the warning about it not being secure.

dirkk0:
ok, I see - so we have 4 possible levels:
- no certs
- certs with mkcert
- certs with letsencrypt
- production setup with nginx et al
which we should somehow document, right?

vincentfretin:
yes

PR #259, opened 2021-03-06
Remove browserify

### LINKED ISSUE #252: Remove browserify?
As far as I understand browserify is used only for the `npm run build` command to create `examples/build.js` (that is executed by `npm run dev`)
I copied the webpack-dev-middleware section from `server/socketio-server.js` (previously `server/index.js`) to `server/easyrtc-server.js` but I'm not sure this is currently used.
I think we need to modify the examples to use
```
<script src="/dist/networked-aframe.js"></script>
```
instead of
```
<script src="/build.js"></script>
```
for it to work no?
If this is the case we can remove the browserify dependencies and the build command.

vincentfretin:
Remove browserify and use only webpack for the examples.
This closes #252


vincentfretin:
About webpack-dev-middleware:
https://webpack.js.org/guides/development/#using-webpack-dev-middleware
Actually it seems @haydenjameslee tried to use it, but the changes weren't finished. :)

vincentfretin:
If anyone want to double check what I did, please take a look.

vincentfretin:
I'm building now `/dist/networked-aframe.js` when running `npm start` because some articles out there are using `npm start` after cloning the repo, and it currently uses the old released version 0.7.1 without easyrtc adapter. But it will use a more recent version if you executed `npm run build` before calling `npm start`. This may be super confusing! I removed the `window.NAF || document.write` hack that was doing this.

vincentfretin:
And I added a note about getting latest networked-aframe release or build.
I'm merging it.

Issue #260, opened 2021-03-07
Add Streaming Video


kylebakerio:
I have an initial implementation with the open-easyrtc adapter working locally that requires just a bit of updating to add here. It follows the semantic pattern of the `networked-audio-source` and its usage pattern. I should have some time to add it in early April.

In the future there is interest in adding multi-video/multi-audio support. In the [w3 spec](https://www.w3.org/TR/mediacapture-streams/#stream-api), it's indicated that this should be possible by adding multiple tracks to one stream, and then extracting the desired tracks as needed. That would be the lowest effort and cleanest implementation. However, it's possible that the spec isn't followed well here--it doesn't seem it was in the past. I plan to try a low-effort experiment to see if this works. If need, we can start looking at the updates needed to support multiple active mediastreams per user.

vincentfretin:
@hthetiot did a great PR there #261 implementing a single video per user. Let's first review that.

hthetiot:
I would recommend to close this issue and create separate issue for multistream.

Then i will add it to adapter based on janus adapter interface and ads sample with screenshare+webcam.

vincentfretin:
ok, see #269

Issue #211, opened 2020-06-13
video streams?


kylebakerio:
Does this support webrtc video streams, or only audio?

edit: looks like only audio.

I'd like to get webrtc video streams.

How hard would it be to add this feature?

craftecode:
This example on Hayden's Glitch might be relevant - https://glitch.com/~karaoke-nite

kylebakerio:
Sadly, every user streams it themselves from a plain old URL, so it doesn't really help me. I already have every user streaming themselves to themselves from their own webcam and it displaying. The part I am unsure about it accessing the webRTC connection and passing video along it.

kylebakerio:
So, I've gone ahead and (very roughly) hacked in the ability to do this. It's pretty much just a proof of concept, but I do have it up and running. Several things are not up to par, but I feel like with just a bit of cleaning up, it could be mergable. Just, this is very organized code, and I don't want to stop on toes and could use some guidance on what kind of tests I should be adding.

https://github.com/kylebakerio/networked-aframe

Still needs some glitches worked out. I was able to get two-way-video working, but usually only the second user gets to see the video of the first user, in my test right now. The current design of the code requires pre-adding a video element with id "webrtc-screen" in the 'assets' section in advance. That just suited me, though, and I'm not sure what the best way to go about this would be. I saw audio elements are dynamically added, but that's a bit different...

The other thing is I haven't really tested with more than two users right now (that's all my app is designed for), so I'll need to try that out as well.

Strange thing I noticed: my audio became suuuuuper laggy, like 15 seconds delayed, while the video stayed crisp and in sync. Probably just a webrtc issue, not sure what could be done in code to fix this, or whether that had anything to do with what I changed (I doubt it, as I mostly left the existing audio code in tact).

@haydenjameslee thoughts?

kylebakerio:
Just adding for completeness, this is likely due to underlying problems in the webrtc adapter of NAF, as discussed [here](https://github.com/networked-aframe/networked-aframe/issues/217). The changes I made in my fork, while incomplete from an API point of view and hacky from a code quality point of view, should basically be what is needed at a fundamental level to add video.

[...]

Issue #190, opened 2020-04-30
Let application share/stream video to other peers


djoksimo:
Hi 👋 ,

Firstly, I just wanted to say that `networked-aframe` is really cool! 😄 

I've scoured through the NAF documentation, PRs and issues, but I'm having trouble setting up video sharing/streaming via the `webRTC` adapter.

I found this Issue https://github.com/networked-aframe/networked-aframe/issues/57 (for EasyWebRTC, which is deprecated) and I found this line of code: https://github.com/networked-aframe/networked-aframe/blob/4c4c1e4f790e68ce0ceab4bbc575c596aebb1a00/src/adapters/naf-webrtc-adapter.js#L316 which says `WebrtcAdapter does not support video yet."`.

Is video chat possible with the NAF `webrtc` adapter (or some other alternative)?

I have the following template:

```html
        <template id="video-template">
          <a-entity class="webcam-entity">
            <video class="vid" id="webcam" playsinline></video>
            <a-entity
              geometry="primitive: circle; radius: 3"
              class="vid-entity"
              random-position
              rotation="0 45 0"
              shadow
              material="src: #webcam"
            ></a-entity>
          </a-entity>
        </template>

...

      <a-entity networked="template:#video-template"></a-entity>

```

haydenjameslee:
Its not supported yet but its certainly possible. The Mozilla Hubs team have implemented WebRTC video into their fork of networked-aframe so there's the existence proof. I'm not aware of any adapters aside from their's that has implemented it however, and i'm not sure how easy their's would be to use. It wouldn't be too hard to add it to the webrtc adapter yourself however, just following a simple tutorial like this: https://www.scaledrone.com/blog/webrtc-tutorial-simple-video-chat/

kylebakerio:
I had the same issue and came up with a (currently crude) example: https://github.com/kylebakerio/networked-aframe

hthetiot:
@djoksimo Need need to wait next April to me done, here is the PR https://github.com/networked-aframe/networked-aframe/pull/261

Issue #55, opened 2017-11-10
In networked-aframe-audio, avatars aren't visible until microphone permission is granted


OrenShalev:
When first entering a new host serving the audio demo, the avatar isn't visible. The last message in the browser console is "Networked-Aframe Connecting..."
When allowing the host access to the microphone, the message "Networked-Aframe Client ID: MXSscnIe..." appears and the avatar becomes visible.

I suggest to make the avatars visible immediately, even if the user did not allow mic access, to provide users an interesting experience as soon as they enter the demo page.

Used Chrome on OS X and Chrome on Win7, networked-aframe-audio.glitch.me currently using  networked-aframe@^0.4.0.

vincentfretin:
There was a fix for the webrtc adapter here https://github.com/networked-aframe/networked-aframe/pull/201 to allow to connect without the mic. We may add a similar fix for the easyrtc adapter, PR welcome.

PR #261, opened 2021-03-08
Add Streaming Video support with example

hthetiot:
- [x] Implement `networked-scene` video options from component element
- [x] Add `enableVideo` argument to `connect` method in `NetworkConnection`
- [x] Implement `EasyRtcAdapter` video support and `setLocalMediaStream` and `enableMicrophone` like Janus Adapter.
- [x] Implement `basic-video` example

Screenshot 2021-03-08 at 03 12 34

Note: Camera from OBS Virtual Camera in screenshot.


hthetiot:
Related: #261 #211 #18 #190

vincentfretin:
Thanks @hthetiot for this contribution. I looked at it quickly, the changes are really great and reflect the discussion we had here. I'll take a second look later to a do a proper review.

examples/basic-video.html
```diff
@@ -0,0 +1,101 @@
...
+        <img id="grid" src="https://img.gs/bbdkhfbzkk/stretch/https://i.imgur.com/25P1geh.png" crossorigin="anonymous">
+        <img id="sky" src="https://i.imgur.com/WqlqEkq.jpg" crossorigin="anonymous" />
+
+        <!-- Templates -->
+
+        <!-- Avatar -->
+        <template id="avatar-template">
+          <a-entity class="avatar">
+             <a-plane  color="#FFF" width="4" height="3" position="0 .6 0"  material="side: front" networked-video-source></a-plane>
+             <a-plane color="#FFF" width="4" height="3" position="0 .6 0"  material="side: back" networked-video-source></a-plane>
```
vincentfretin, line 35: Why two planes? Doesn't `side:both` work?

examples/basic-video.html
```diff
@@ -0,0 +1,101 @@
...
+    <script src="https://unpkg.com/aframe-particle-system-component@1.0.5/dist/aframe-particle-system-component.min.js"></script>
+    <script src="/js/spawn-in-circle.component.js"></script>
+  </head>
+  <body>
+    <a-scene networked-scene="
+      room: basic-video;
+      debug: true;
+      adapter: easyrtc;
+      audio: false;
+      video: true;
```
vincentfretin, line 22: Actually I was more leaning to remove the audio and video options from networked-scene, see discussion in https://github.com/networked-aframe/networked-aframe/issues/80#issuecomment-360659707

src/adapters/EasyRtcAdapter.js
```diff
@@ -31,16 +31,17 @@ class EasyRtcAdapter extends NoOpAdapter {
...
 
-    this.easyrtc.enableVideo(false);
+    this.easyrtc.enableVideo(options.video);
     this.easyrtc.enableAudio(options.audio);
 
-    this.easyrtc.enableVideoReceive(false);
-    this.easyrtc.enableAudioReceive(true);
+    // TODO receive(audio|video) options ?
+    this.easyrtc.enableVideoReceive(options.video);
+    this.easyrtc.enableAudioReceive(options.audio);
```
vincentfretin, line 44: See https://github.com/networked-aframe/networked-aframe/commit/fccad7543f1255385ec0c9097044b39ab09540b2 were we changed from `options.audio` to `true`. So it seems you will introduce a regression here.

src/adapters/EasyRtcAdapter.js
```diff
@@ -177,19 +173,73 @@ class EasyRtcAdapter extends NoOpAdapter {
...
+          this.pendingMediaRequests.get(clientId).video = { resolve, reject };
+        });
+
+        this.pendingMediaRequests.get(clientId).audio.promise = audioPromise;
+        this.pendingMediaRequests.get(clientId).video.promise = videoPromise;
+
+        audioPromise.catch(e => NAF.log.warn(`${clientId} getMediaStream Audio Error`, e));
+        videoPromise.catch(e => NAF.log.warn(`${clientId} getMediaStream Video Error`, e));
+      }
+      return this.pendingMediaRequests.get(clientId)[type].promise;
```
vincentfretin, line 198: I recognize the code from the janus adapter and I know this code works well. :)

src/components/networked-video-source.js
```diff
@@ -0,0 +1,87 @@
...
+      if (newStream) {
+        this.video.srcObject = newStream;
+
+        var playResult = this.video.play();
+        if (playResult instanceof Promise) {
+          playResult.catch((e) => naf.log.error(`Error play video stream`, e));
+        }
+
+        const mesh = this.el.getObject3D('mesh');
+        mesh.material.map = new THREE.VideoTexture(this.video);
```
vincentfretin, line 51: Fo you want to set `this.videoTexture` here?

src/components/networked-video-source.js
```diff
@@ -0,0 +1,87 @@
...
+  _clearMediaStream() {
+    if (this.video) {
+      this.video.srcObject = null;
+      this.video = null;
+      this.stream = null;
+    }
+  },
+
+  remove: function() {
+    if (!this.videoTexture) return;
```
vincentfretin, line 68: `this.videoTexture` is not currently set.
We may explicitly dispose the texture too. See https://github.com/mozilla/hubs/blob/fef69c13341ec148211c53b3b5d6f2790519ec65/src/components/video-texture-target.js#L105

src/components/networked-video-source.js
```diff
@@ -0,0 +1,87 @@
...
+
+    if(newStream != this.stream) {
+      if (this.stream) {
+        this._clearMediaStream();
+      }
+
+      if (newStream) {
+        this.video.srcObject = newStream;
+
+        var playResult = this.video.play();
```
vincentfretin, line 45: Calling play is necessary if you reuse the video element and set again srcObject I guess. Ok. Otherwise when we create the video element it's with autoplay already.

vincentfretin (changes requested):
This a great first implementation. Thank you for contributing it. I left some comments.
I'm wanting to discuss further the video option here and how we can use setLocalMediaStream.
From my understanding, the mic and camera tracks are set by the open-easyrtc library and the stream acceptor will call setMediaStream with a stream object containing the tracks. And we will have an audio or video track depending on enableMicrophone and enableCamera.
So I'm not sure if we can use setLocalMediaStream here like we do in janus adapter where we get the mic stream ourself and call setLocalMediaStream and this will change the stream sent by the publisher RTCPeerConnection, see https://github.com/Synantoo/naf-janus-adapter/blob/a3fbd137bebf3e4d34b4ba11a9ce91a6985cd477/examples/index.html#L126-L130 and https://github.com/Synantoo/naf-janus-adapter/blob/261e1ca6040b74c4b90ba62762e0847ab8b76761/src/index.js#L914 
This let you create your own ui to list available microphones, get the mic stream with getUserMedia and an exact deviceId match. In janus adapter setLocalMediaStream is using replaceTrack on the stream of the publisher RTCPeerConnection.
In your implementation, setLocalMediaStream doesn't seem to change anything on the RTCPeerConnection, so it seems that if you select another mic and call setLocalMediaStream, it will still send the audio track selected by open-easyrtc. Am I wrong?

src/adapters/EasyRtcAdapter.js
```diff
@@ -177,19 +173,73 @@ class EasyRtcAdapter extends NoOpAdapter {
...
+        stream
+    );
+  }
+
+  enableMicrophone(enabled) {
+    this.easyrtc.enableMicrophone(enabled);
+  }
+
+  enableCamera(enabled) {
+    this.easyrtc.enableCamera(enabled);
```
vincentfretin, line 240: Maybe calling ` this.easyrtc.enableVideo(true)` make sense here before calling enableCamera? This is to renegotiate the sdp offer right?

src/adapters/EasyRtcAdapter.js
```diff
@@ -31,16 +31,17 @@ class EasyRtcAdapter extends NoOpAdapter {
...
 
-  // options: { datachannel: bool, audio: bool }
+  // options: { datachannel: bool, audio: bool, video: bool }
   setWebRtcOptions(options) {
     // this.easyrtc.enableDebug(true);
     this.easyrtc.enableDataChannels(options.datachannel);
 
-    this.easyrtc.enableVideo(false);
+    this.easyrtc.enableVideo(options.video);
     this.easyrtc.enableAudio(options.audio);
```
vincentfretin, line 40: If we don't use enableVideo and enabledAudio here, we may still connect without the mic if permission was denied? see #55

src/adapters/EasyRtcAdapter.js
```diff
@@ -177,19 +173,73 @@ class EasyRtcAdapter extends NoOpAdapter {
...
+  }
+
+  setLocalMediaStream(stream) {
+    this.setMediaStream(
+        this.easyrtc.myEasyrtcid,
+        stream
+    );
+  }
+
+  enableMicrophone(enabled) {
```
vincentfretin, line 235: Call `this.easyrtc.enableAudio(enabled)` here?

src/components/networked-video-source.js
```diff
@@ -0,0 +1,87 @@
...
+      this.setupVideo();
+    }
+
+    if(newStream != this.stream) {
+      if (this.stream) {
+        this._clearMediaStream();
+      }
+
+      if (newStream) {
+        this.video.srcObject = newStream;
```
vincentfretin, line 43: Maybe create a new stream with explicitly only the first video track before setting srcObject?
Not tested:
```
newStream = new MediaStream(newStream.getVideoTracks()[0]);
this.video.srcObject = newStream;
```
maybe make sure we have a least a video track.

vincentfretin:
If we want to stream both camera and a screenshare, one need to call `register3rdPartyLocalMediaStream` and `addStreamToCall` like you said in https://github.com/networked-aframe/networked-aframe/issues/211#issuecomment-792297367
Is there something that need to be done for the other participants on the receiving part? Do we have several videos tracks then in `this.mediaStreams[clientId]["video"]` automatically? Do we know if the order is kept? Are we sure first video is the camera and second video is a screenshare? What if we call `enableCamera(false)` and keep the screenshare?

vincentfretin:
You can also add a link from index.html to your basic-video example.

examples/basic-video.html
hthetiot: I never used aframe before that why :) Will do.

examples/basic-video.html
hthetiot: Well, i think that important to be able to do audio or/and video only.
Can we do that in a separate PR?

src/adapters/EasyRtcAdapter.js
hthetiot: ok will set default to true.

src/components/networked-video-source.js
hthetiot: yeah was not sure, let do that.

src/components/networked-video-source.js
hthetiot: will do.

src/components/networked-video-source.js
hthetiot: yes this also allow better auto-play support under certain cases if autoplay failed and user interacted then it will play.

examples/basic-video.html
vincentfretin: That's fine with me.

hthetiot:
> You can also add a link from index.html to your basic-video example.

done

examples/basic-video.html
hthetiot: I have tried but it does not work unfortunately.

src/components/networked-video-source.js
hthetiot: I already use "NAF.connection.adapter.getMediaStream(ownerId, "video")" with video parameter for that and call _setMediaStream. if null it will clear previous if there was any, otherwise do nothing and set stream to null.

src/adapters/EasyRtcAdapter.js
hthetiot: no, that on the _connect method of the adapter, i have not changed that behavior will be the same as before with audio, if denied will not connect. 

Original: https://github.com/networked-aframe/networked-aframe/blob/master/src/adapters/EasyRtcAdapter.js#L219
New: https://github.com/networked-aframe/networked-aframe/blob/db099d8396113b83f9cd256b1a4e7115c175fd45/src/adapters/EasyRtcAdapter.js#L260

hthetiot:
> f we want to stream both camera and a screenshare, one need to call `register3rdPartyLocalMediaStream` and `addStreamToCall` like you said in [#211 (comment)](https://github.com/networked-aframe/networked-aframe/issues/211#issuecomment-792297367)
> Is there something that need to be done for the other participants on the receiving part? Do we have several videos tracks then in `this.mediaStreams[clientId]["video"]` automatically? Do we know if the order is kept? Are we sure first video is the camera and second video is a screenshare? What if we call `enableCamera(false)` and keep the screenshare?

Multi stream is more complicated, we need change on Adapters interface for that. Let keep that for another PR dedicated to multi-stream.

hthetiot:
> So I'm not sure if we can use setLocalMediaStream here like we do in janus adapter where we get the mic stream ourself and call setLocalMediaStream and this will change the stream sent by the publisher RTCPeerConnection, see https://github.com/Synantoo/naf-janus-adapter/blob/a3fbd137bebf3e4d34b4ba11a9ce91a6985cd477/examples/index.html#L126-L130 and https://github.com/Synantoo/naf-janus-adapter/blob/261e1ca6040b74c4b90ba62762e0847ab8b76761/src/index.js#L914
> This let you create your own ui to list available microphones, get the mic stream with getUserMedia and an exact deviceId match. In janus adapter setLocalMediaStream is using replaceTrack on the stream of the publisher RTCPeerConnection.
> In your implementation, setLocalMediaStream doesn't seem to change anything on the RTCPeerConnection, so it seems that if you select another mic and call setLocalMediaStream, it will still send the audio track selected by open-easyrtc. Am I wrong?

I would prefer to do that in separate PR, before or may be while doing multistream.

hthetiot:
> What if we call `enableCamera(false)` and keep the screenshare?

No by enableCamera will, again that for multi-stream next PR, will also do multi-stream sample when i do that anyway.

vincentfretin:
I'm fine with doing more complex stuff in another PR. In this case I think you should remove the setLocalMediaStream api from this PR is it doesn't do what I expect.

hthetiot:
> setLocalMediaStream

Done, I will do the equivalent of setLocalMediaStream using easyrtc in a separate PR later.

vincentfretin (approved):
The latest changes are good.
I tried running the example right now, I didn't until now.
I configured a certificate in `server/easyrtc-server.js` as documented in #257 and ran `npm run dev`
The basic-audio example works fine.
With your basic-video, I'm prompted for the camera permission, good, but I get the error in the console for Firefox
"MEDIA_ERR Failed to get access to local media. Error code was AbortError."
In Chrome I have no error, but in both browsers I don't see the <a-entity> element created for the other participant.
If now I set networked-scene="video:false", I see the white plane of the other participant. If now I call `NAF.connection.adapter.enableCamera(true)` is does nothing, no error.
I'll test again tomorrow after a reboot, maybe there is something weird with my pc.
But in any case, can we make sure we still fully connect if there is an error getting the camera?

You don't have to push the dist folder in the PR. I'll take care of pushing an up to date dist folder after the merge. You can rebase your branch to remove those "build dist" commits if you want. The `npm run dev` command take care of building dist/networked-aframe.js in memory via webpack-dev-middleware. And `npm start` will build them on the filesystem (just fixed an issue with that on master after I merged the "remove browserify" PR, it seems I didn't retested properly the latest commit I did on that, sorry)

vincentfretin:
Nope, after a reboot I have the same issue I described. Is this working properly on your side?
Even if we don't do multi video streams in this PR, wouldn't it be good to have a mode where we share the screen/window/tab instead of the camera as part of this PR? We may think the api now before merging this. Maybe `NAF.connection.adapter..enableVideo(true, type="camera")` instead of `NAF.connection.adapter.enableCamera(true)`? And we could call `NAF.connection.adapter.enableVideo(true, type="screen")` to trigger screen sharing instead of the camera?

hthetiot:
> "MEDIA_ERR Failed to get access to local media. Error code was AbortError."

Occurs when another application use you camera on firefox.

Works for me with Chrome and Firefox.

Screenshot 2021-03-08 at 19 49 04

> NAF.connection.adapter.enableCamera(true)

Will require to call initMedia again, enableCamera and enableMicrophone can be use to disable media stream not create media stream.


> You don't have to push the dist folder in the PR.

Ok noted.

> You can rebase your branch to remove those "build dist" commits if you want. 

I rather not please, merge and rebuild, next time i will not make dist.

> The npm run dev command take care of building dist/networked-aframe.js in memory via webpack-dev-middleware.

I already use `npm run dev`, i commited dist because i saw it was dont on other PR (e.g https://github.com/networked-aframe/networked-aframe/pull/259)

hthetiot:
> But in any case, can we make sure we still fully connect if there is an error getting the camera?
Works for le on MacOS FF, Chrome Safari and Windows on Chrome, Edge.
I never push unless it's tested, if you are on Linux you may have your own issues.

Screenshot 2021-03-08 at 19 58 16

hthetiot:
> Even if we don't do multi video streams in this PR, wouldn't it be good to have a mode where we share the screen/window/tab instead of the camera as part of this PR? 

NO this is another feature. Trust me it's better to go step by steps.

hthetiot:
> NAF.connection.adapter.enableVideo(true, type="camera") instead of NAF.connection.adapter.enableCamera(true)? And we could call NAF.connection.adapter.enableVideo(true, type="screen") to trigger screen sharing instead of the camera?

That not what Janus adapter does.
Trust me the more you add the more you going to have problems to tests, and the current networked-aframe API need to be thought before doing multi-stream or screenshare. EasyRTC does not provide screen share it only expose reigster custom stream.

Please dont ask again, i dont like when people want feature A and then ask B and C when its not in the same scope and A is already fulfilling  #260 #211 #18 #190 scope. Feature PR need to be progressive the more you add the more you have risk of breking and bug, beside setLocalMediaStream this match janus adapter.

hthetiot:
you need to provide what "aframe" api will look like, you already want to remove audio: true, video: true

You dont realize that multi stream will require negotiation and special code when the caller have more than one stream due WebRTC bugs, we will need to send the first stream, then once establish send the nexts ones and negotiate.

Video is step one, then if you provide proper instruction for interface, like you did on #211  i may do more. But again mixing multi stream and video stream in one PR is a mistake, it better to iterate than making big PR that change everything for no good and have review that never ends.

<3

hthetiot:
I used `enableCamera` to match janus adapter `enableMicrophone` logic. I'm open to change that minor.

hthetiot:
Edge ok, but will result into error "MEDIA_ERR Failed to get access to local media. Error code was NotReadableError." if another browser use the camera, similar to linux AbortError. Check slack PM for private test URL i sent you.

hthetiot:
Ok on Safari iOS 14.4 and Android Chrome and all combination Windows (Edge/Chrome <-> Safari <-> Firefox <-> Android <-> iOS.

Screenshot_20210308-205106_Chrome

hthetiot:
> I understand the multi streams videos is a complex subject. Here I was talking about screensharing in replacement of the camera, so not multistream, just one stream. But it seems you can't easily do that with open-easyrtc api, compared to setLocalMediaStream that the janus adapter api propose, noted.

Yes, replacing stream is actualey a subset of multi-stream (Transceiver/Sender related). Still i can do `setLocalMediaStream` in one step then once the API for multistream is defined i can do Multi-stream.

It can be done with open-easyrtc today, my issue is more what the API will look like on networked-aframe to add a stream. may be it can simply be `addLocalMediaStream(stream, streamName)` and `removeLocalMediaStream(stream, streamName)`  not sure what you d'like.

hthetiot:
>  Hum I thought this.easyrtc.enableVideo(enabled); would create it, maybe it does? 
No it should only disable if sending one depending the browsers involved it may freeze, get black or blank.

examples/basic-video.html
hthetiot: Cycle and cube geometry works i tested

hthetiot: Not plane

PR #264, opened 2021-03-24
check for empty occupants fixed

czadamToptal:
For me that check always returned false and the room never got cleaned up properly

### DIFF
--- a/server/socketio-server.js
+++ b/server/socketio-server.js
@@ -76,7 +76,7 @@ io.on("connection", socket => {
       const occupants = rooms[curRoom].occupants;
       socket.to(curRoom).broadcast.emit("occupantsChanged", { occupants });
 
-      if (occupants == {}) {
+      if (JSON.stringify(occupants) === JSON.stringify({})) {
         console.log("everybody left room");
         delete rooms[curRoom];
       }


server/socketio-server.js
```diff
@@ -76,7 +76,7 @@ io.on("connection", socket => {
       const occupants = rooms[curRoom].occupants;
       socket.to(curRoom).broadcast.emit("occupantsChanged", { occupants });
 
-      if (occupants == {}) {
+      if (JSON.stringify(occupants) === JSON.stringify({})) {
```
vincentfretin, line 79: You're right
```
occupants={}
occupants == {}
false
```

Instead of jsonify to do a comparison.
`if (Object.keys(occupants).length === 0)`
should be better, this is what open-easyrtc more or less do via the underscore size function.
Can you please test that and tell me if it's good for you?

vincentfretin:
Thank you for the contribution! If you can test the change I proposed and update the PR, it would be great.
Otherwise I will test that myself later.

server/socketio-server.js
czadamToptal: `if (Object.keys(occupants).length === 0)
`
works as well

vincentfretin:
Thanks I fixed it in https://github.com/networked-aframe/networked-aframe/commit/6606e1462b8ef3c2dbde4362df4dd81dea6cf0c9

Issue #273, opened 2021-04-17
Update or remove snow particles in the examples


vincentfretin:
I just found out that the shader error we have in the examples is related to the snow particles. It worked on aframe 1.0.4, but doesn't work anymore on aframe 1.1.0 and 1.2.0.
There is an issue about it https://github.com/IdeaSpaceVR/aframe-particle-system-component/issues/59 and PR https://github.com/IdeaSpaceVR/aframe-particle-system-component/pull/60

You can replace
```
<script src="https://unpkg.com/aframe-particle-system-component@1.0.5/dist/aframe-particle-system-component.min.js"></script>
```
by
```
<script src="https://cdn.jsdelivr.net/gh/oneWaveAdrian/aframe-particle-system-component@aframe-1.2.0-upgrade/dist/aframe-particle-system-component.min.js"></script>
```
and the snow will work again with aframe 1.2.0.

I'm wondering if we wait for the PR to be merged so we can use a new release of it, or just remove this from the examples so we have one less thing to maintain?

vincentfretin:
The url above doesn't work anymore because the branch was merged into IdeaSpaceVR fork and the branch deleted.
You can use now:
```
<script src="https://cdn.jsdelivr.net/gh/IdeaSpaceVR/aframe-particle-system-component@master/dist/aframe-particle-system-component.min.js"></script>
```

Issue #267, opened 2021-03-27
Schema may not be registered when the networked component is initialized


vincentfretin:
There can be a race condition case where the schema defined in the footer of the page is not register when the networked init function is executed where it calls `NAF.schemas.getComponents` to get the components to sync based on the registered template
https://github.com/networked-aframe/networked-aframe/blob/393ec7b8ffe22ff20e4b2f84fe6f32a7c23f7c06/src/components/networked.js#L154
The issue seems to be always reproducible when there is no assets to download.
I created an example based on the naf-project template where I only commented the two assets sky and grid, and modified the template to not use random-color, which hide the issue actually...

https://glitch.com/edit/#!/naf-schema-race-condition

The issue here is that all avatars stay red, instead of a random color set by the participant and normally synced to other participants.
In the logs we get this order:
- calling NAF.schemas.getComponents from networked component init
- components in schema:  ["position", "rotation"]
- register schema
- scene loaded

With this order, `NAF.schemas.getComponents` returns the default `["position", "rotation"]` instead of `["position", "rotation", {selector: ".head", component: "material", property: "color"}]`, so not syncing the material color set by the random-color component on the player entity.

A workaround is to not set the networked component in the html, but set it with javascript:
```
     <script>
      document.addEventListener('DOMContentLoaded', () => {                           
        const scene = document.querySelector('a-scene');
        const sceneLoaded = () => {
          console.log("scene loaded");
          document.getElementById("player").setAttribute("networked", "template:#avatar-template;attachTemplateToLocal:false;");
        }
        if (scene.hasLoaded) {                                                         
          sceneLoaded();
        } else {
          scene.addEventListener('loaded', sceneLoaded);
        }
      });
    </script>
```
and remove `networked="template:#avatar-template;attachTemplateToLocal:false;"` from `<a-entity id="player"`

I'm not sure how we can fix that... or if we can fix that really...

The issue may appear randomly when you have small asset in `<a-assets>` to download or if you have it already in your cache.
The issue is much worse when you register a template that doesn't have position and rotation, you can end up with two participants having a different schema, one sending a vec3 trying to fit as a number in a custom component and you have weird things happening like objects disappearing. Yes I had this case on one of my projects. :D

vincentfretin:
[...]

PR #254, opened 2021-02-27
Fix the a-saturday-night example with latest aframe 

vincentfretin:
This is half fixed, may need to retrieve new assets from https://aframe.io/a-saturday-night/
https://github.com/aframevr/a-saturday-night

This closes #239

Issue #233, opened 2021-01-19
hand-controls


ruliusuifeng:
Hello.
When I synchronized the model of the hand, in my perspective, I manually moved other people's hands along with it.
 I want us to wave our hands differently, for example, let the other person see me compare Thumbs up

vincentfretin:
Sorry I don't fully understand what you are saying. I don't know if you have an issue or want an enhancement.
Does the examples/tracked-controls.html works for you? What are you trying to achieve compared to this example? Can you share an example on glitch.com?
Ross here
https://aframevr.slack.com/archives/C65MC3NSF/p1613982391032200?thread_ts=1613920593.023200&cid=C65MC3NSF
wanted something like synchronized basic hand gesture when detecting the different pressed buttons like grip gesture, this is what you are referring with Thumbs up?

kylebakerio:
@vincentfretin I also wanted this feature so today decided to hack it out. I did 98% of the work, just have some little clean up tweaks to make: https://nctl2.glitch.me/tracked-controllers.html

Feel free to test it and let me know what you think, I'll make a pull request soon.

vincentfretin:
Great, I need to test that!

vincentfretin:
#355 is addressing it.

PR #290, opened 2021-08-15
Tracked Controllers Example Overhaul

kylebakerio:
- Updated environment to be more appealing, and suggest that we update all updated examples to this environment both for visual appeal, but also to make it clear that these new examples are separate from the old examples. 
- Removed particles.
- Made tracked hands actual hand models.
- Added basic illustration of how to add hands dynamically.
- Added comments.
- Removed unnecessary schema examples because position and rotation are tracked by default.
- Added hand model assets.
- Removed lighting,sky,grid,ground--all now handled by environment component, making the example less distracting.
- Changed avatar template code style to be more visually clear (no need to add emphasis to the avatar geometry with newlines).
- Changed avatar proportions so that hands are not swallowed by them, and changed proportions slightly to make them somewhat less dorky looking--I also suggest this modification across all examples to create a new coherent visual style for current examples, although I'm open to better options.
- Added joystick movement controls and flying ability via aframe-extras.
- Added comments mentioning room for future development.

I'd like to go through the examples and update all avatars and environments and examples to match the various changes I've made here, if you're comfortable with that.

Also, would be good to share editing rights with me on glitch so I can make the corresponding updates there to the canonical examples.

You can try this out here:
https://glitch.com/edit/#!/aug2021-naf?path=examples%2Ftracked-controllers.html%3A42%3A0


vincentfretin:
Thanks for starting to modernize the tracked controllers example. The hands tracking definitely didn't exist when this example was created, this has since been implemented in WebXR and included in latest aframe, so it will be a nice addition. There is also a demand for this kind of example #233.

First I'll share my thoughts on the examples and the maintenance of them, then I'll comment of some specific points of your list.
Note that I didn't test your example, I don't have currently the Quest with me.

As a maintenance point of view of this repository, I prefer we depend less on external libraries in the basic examples. I asked about removing particles dependency in the examples in #273, in the end I didn't want to change drastically the examples, so I just updated the url to a fork. I'm open to remove it completely and keep the examples simpler. Similar to that, in the end I also removed the a-saturday-night example (#254) because we couldn't update it to latest aframe. So we need to be careful of external dependencies we use in the examples.
Although the aframe environment is great and may be more appealing, I think using the aframe primitives a-sky, the plane, lights and declaring the assets may be better for educational purpose, and the entities can be easily modified. Also I may be nostalgic of the tron like environment if we remove it. :)
I'm not against a more complex and appealing example though, we can have an example with hands like your PR combined with physics. We can even combine what I started in #270 with yours and use the aframe environment component for this one.

You're right about the [schema workaround](https://github.com/networked-aframe/networked-aframe/issues/267#issuecomment-839712517), we need to put it in all examples now.
For glitch, I never had access to the old examples. I only have access to the [naf-project](https://glitch.com/edit/#!/naf-project) I created, and I already did the schema workaround on this one.

I think the tracked-controllers example is the only one where we use nested networked entities, although [position, rotation] is the default schema, I think it's ok to have it explicitly. Having them already defined is nice if you decide to modify the example and add an additional component you want to sync to your left arm. :)
I see the examples as more educational examples where the developer can build upon.
I often compared two basic examples to see what lines needs to be changed to support a given use case, so I definitely prefer that the examples stay basic and that we have the same html indentation between them. In my opinion having all attributes of an html tag on the same line make it more difficult to compare. Also my personal preference is to have an attribute per line, it's way better with `git diff` :D

So in summary I'd prefer we update the tracked controllers with the new hands model, but without the aframe-extras and environment if possible, and with the same indentation that we have in the other examples.
Then we can add a new full example with environment, aframe-extras with animations with animation-mixer, even physics.
What do you think?

kylebakerio:
I'd like to expand this to a component that could transmit specific controller state, like is requested in #233, and plan to implement that at some point for my own app as well. At some point, we'll want to actually include hand-tracking, as in a full networked hand skeleton, but both of those are milestones that follow after this one, I think.

Dependies: That makes sense. I generally agree. I think that movement-control is pretty standard and a far more valuable (allows experiencing networked movement a bit more) and reliable dependency than a particle component, though. I'd suggest we keep it and add it to all examples, but I'm open to keeping it out.

I do think the particles look a bit cheesy, though, and would prefer we didn't keep those.

> Although the aframe environment is great and may be more appealing, I think using the aframe primitives a-sky, the plane, lights and declaring the assets may be better for educational purpose, and the entities can be easily modified. Also I may be nostalgic of the tron like environment if we remove it. :)

My thoughts are that they are distracting to newcomers. I think it's a significant amount of extra code that has nothing to do with NAF. I'd rather we educate users on NAF-specific things (e.g. movement-controls inclusion) than general A-Frame things don't affect NAF. To that point, the environment component makes the code for the environment be completely undistracting, while also more appealing.

I think the environment looks quite out of date, personally. It reminds me of the old A-Frame 0.3 examples. It's also too dark. But if you prefer Tron, there's a Tron preset in the environment component as well, which I can switch to. I guess that would have more continuity with some kind of 'tradition'? :)

> You're right about the schema workaround, we need to put it in all examples now.
Ideally we fix the issue so that we have a better, consistent API that doesn't look like a hack, long term. We should probably fix that before we go inject the workaround to all the examples, only to have to update them all again afterwards.

> For glitch, I never had access to the old examples. I only have access to the naf-project I created, and I already did the schema workaround on this one.
Sure, I just mean the glitch referred to in the readme now. I assume we're just using the one you created as the official one now.

> I often compared two basic examples to see what lines needs to be changed to support a given use case, so I definitely prefer that the examples stay basic and that we have the same html indentation between them. In my opinion having all attributes of an html tag on the same line make it more difficult to compare. Also my personal preference is to have an attribute per line, it's way better with git diff :D

In general I understand and agree, however, I think in this case, that huge amount of indentation and so many lines makes it feel visually overwhelming and draws a lot of focus to parts of the code that don't matter. It's much easier to see the structure of what is going on with this style than with a bunch of custom geometry and material attributes laid out to craft a 3d model for an avatar in code. In 'real' projects, I definitely separate out the attributes per-line as well, I just think that pushing those details aside makes for better NAF educational content. I feel really strongly about this; the way the original examples show nested components with all those lines of attributes makes it really hard to absorb what is happening, in my opinion. I'd say if you insist on not changing to this style, then we should consider just making the avatars far simpler--the way they are now is a real problem, I think.

> I think the tracked-controllers example is the only one where we use nested networked entities, although [position, rotation] is the default schema, I think it's ok to have it explicitly. Having them already defined is nice if you decide to modify the example and add an additional component you want to sync to your left arm. :)

I was a bit undecided on this. On the one hand, I don't want to make users think those need to be explicitly decalred when they aren't needed. On the other hand, I agree, it is nice to see how it works. I think I like the idea of leaving it there commented out, with it explained that we could declare it but it would be redundant?

> Then we can add a new full example with environment, aframe-extras with animations with animation-mixer, even physics.
I think it makes sense to add animations /animation-mixer to this one, eventually, by creating a 'networked-hand-controls' component that syncs hand animation across devices, which I'm happy to do. When we do the physics example, I think we should probably keep the animations out of it (unless they're abstracted away to a networked-hand-controls component), and focus on the minimum to get networked physics working--probably just something like having the ability to knock towers of cubes over.

Then, _separately_, we can have a more advanced physics example that incorporates super-hands, ability to grab things, etc.?

kylebakerio:
I've updated the environment to a tron-similar preset that I hope is a decent compromise, added comments showing how the other networked components can be optionally declared but also showing that because of the defaults they aren't needed, and added many other educational code comments to the existing code, and also cleaned up the component a bit.

kylebakerio:
I have to say, the way the sphere is declared with visible:false both in the schema but also in the scene itself is pretty confusing for a basic example. It is interesting to show the specific case of modifying attributes that aren't visible on your own local copy of a template, but it seems like it would be more intuitive to have that `#head` sphere added by the template and then be modified in JS after it is instantiated.

It's just kind of a confusing bit of code to use for an educational demo. I've never used this kind of functionality. I kind of feel like this old-school example should only exist in a single demo showing this concept, and the other demos should just have simple gltf avatars (which are more applicable to common use case) and text usernames tracked under the avatars instead of relying on random-color.

Thoughts?

kylebakerio:
I also feel like having onConnect be called if the global is present is just a terrible API and that should definitely be moved to an event or an exposed promise on the NAF global... I cringe every time I see that.

kylebakerio:
https://8-2021-naf.glitch.me/tracked-controllers.html

vincentfretin:
I like the tron environment from aframe-environment. Ok, let's use that. :) You can change it in all examples in a separate PR so I can merge just this change for now (remove the snow particles, remove the imgur images, plane, lights).

For the indentation part of the template, for this particular case, I changed my mind, I'm ok with putting all components on the same line, but please don't put two spaces between attributes and before the end of the tag. You can do a separate PR just for this too if you want or in the same as the one above so I can quickly review and merge the changes.

> I have to say, the way the sphere is declared with visible:false both in the schema but also in the scene itself is pretty confusing for a basic example. It is interesting to show the specific case of modifying attributes that aren't visible on your own local copy of a template, but it seems like it would be more intuitive to have that #head sphere added by the template and then be modified in JS after it is instantiated.

This can indeed be difficult to grasp why we do this. The `visible:false` sphere is just there so that the `random-color` component that set `material.color` is synchronized to the other participants. We don't need to create eyes or eventually other parts of the avatar, that's why we use `attachTemplateToLocal:false` and define only the head sphere. If you set `attachTemplateToLocal:true` with `visible:false` does it works correctly? Maybe this would be simpler to understand?
If yes, you can do a PR just to fix that in the examples. I like small PR, easier to review and merge if no further discussion is needed. ;)

> It's just kind of a confusing bit of code to use for an educational demo. I've never used this kind of functionality. I kind of feel like this old-school example should only exist in a single demo showing this concept, and the other demos should just have simple gltf avatars (which are more applicable to common use case) and text usernames tracked under the avatars instead of relying on random-color.

Putting the old example https://glitch.com/edit/#!/naf-nametags in the repo and updating it would be great.

> I also feel like having onConnect be called if the global is present is just a terrible API and that should definitely be moved to an event or an exposed promise on the NAF global... I cringe every time I see that.

Ah ah I also find it weird that a global function named onConnect would be called like that out of the blue. :)
We have this API like forever so we can't really change that without breaking code used in the wild.
This is done here https://github.com/networked-aframe/networked-aframe/blob/f1c62b41a4a13c6d4d6f425eaf14c81ab90e44ff/src/components/networked-scene.js#L35-L37
https://github.com/networked-aframe/networked-aframe/blob/f1c62b41a4a13c6d4d6f425eaf14c81ab90e44ff/src/components/networked-scene.js#L56-L58

The global onConnect is the callback function called on the "connected" event you have on document.body, see
https://github.com/networked-aframe/networked-aframe/blob/f1c62b41a4a13c6d4d6f425eaf14c81ab90e44ff/src/NetworkConnection.js#L60-L68
So you could register your own listener instead of relying on the global onConnect, it's just not documented. And you need to be sure you call your function if you are already connected like it's done above.

kylebakerio:
Great, thanks for the responses!

> I like the tron environment from aframe-environment. Ok, let's use that. :) You can change it in all examples in a separate PR so I can merge just this change for now (remove the snow particles, remove the imgur images, plane, lights).

> For the indentation part of the template, for this particular case, I changed my mind, I'm ok with putting all components on the same line, but please don't put two spaces between attributes and before the end of the tag. You can do a separate PR just for this too if you want or in the same as the one above so I can quickly review and merge the changes.

So, to be clear: my idea with this pull request is to establish a 'new standard' for what our demos should look like. Once we've hammered it all out for this one, I'd like to do another pull request to apply the updates here to all the other demos.

> This can indeed be difficult to grasp why we do this. The visible:false sphere is just there so that the random-color component that set material.color is synchronized to the other participants. We don't need to create eyes or eventually other parts of the avatar, that's why we use attachTemplateToLocal:false and define only the head sphere. If you set attachTemplateToLocal:true with visible:false does it works correctly? Maybe this would be simpler to understand?
If yes, you can do a PR just to fix that in the examples. I like small PR, easier to review and merge if no further discussion is needed. ;)

If it's ok with you, I've started a separate issue to carry on this discussion, as I have more thoughts on this topic, and I think it starts to be bigger scope than this pull request. I've started it here: https://github.com/networked-aframe/networked-aframe/issues/292.

As for how to handle that here... whatever we do in regards to the API going forward for handling this, I think in our demos it would make more sense to update color-randomizer to handle this logic internally and to just be applied to the template directly (and to have logic inside that lets it determine whether it should update the color because it is local, or not update the color because it isn't local/the owner), than to demonstrate doing this kind of thing in HTML, which should imo be discouraged and is just awkward and unclear and hard to follow.

> Putting the old example https://glitch.com/edit/#!/naf-nametags in the repo and updating it would be great.

Agreed, and I'm glad to do this. I'd perhaps like to go as far as updating all of our demos to use nametags, possibly instead of colors, though, instead of having one example for nametags. Perhaps nametags and colors could both be done within one new component that makes clear how one updates NAF template instances that are locally instantiated only (which as far as I can tell, is supposed to be the point of including the color-randomizing concept in the demos.)

> Ah ah I also find it weird that a global function named onConnect would be called like that out of the blue. :)
We have this API like forever so we can't really change that without breaking code used in the wild.
This is done here

Ah! Well, if that functionality already exists, can we switch to presenting the listener-based functionality instead in our demos, and updating our documentation to encourage that pattern? I've started a new issue ont his topic as well, https://github.com/networked-aframe/networked-aframe/issues/293.

vincentfretin:
> I understand--after having used this library for some time, now, and having stared at that code for some time--what is going on, but I think NAF is just really weird here. The idea that you let NAF manage creating and syncing your avatar, and tell it to instantiate its invisible copy... but then the user, separately, outside of NAF, makes a child component that is supposed to just happen to match the template, which NAF just happens to somehow know to listen for (probably because of the class and hierarchy), and changes to that user-created child of the NAF-created (but not attached to local, and invisible) template instance just happen to correctly be propagated?

From your long paragraph, I don't know if you understood it or not. :)
Which components and data of those components that will be synchronized is defined by the networked schema, not the template. You can have some entities with specific components in the template that are not synchronized, if there are not defined in the networked schema.

> Personally, I think it's poor design. Even if this does work, this shouldn't be how it's taught to be done. It would make far more sense to write a component that only sets color if it's the user's local avatar but not a networked avatar--similar to to logic employed in networked-audio-source.

How do you check if this is the local avatar or a remote avatar? In one of my app, I'm doing sometimes the logic in the component instead of the html and I'm using `if (this.el.id=== 'cameraRig') to check if it's the local avatar, this does the job, but this is not something we should encourage on a component in a library, this breaks reusability of the component.


> Whatever we do in regards to the API going forward for handling this, I think in our demos it would make more sense to update color-randomizer to handle this logic internally and to just be applied to the template directly (and to have logic inside that lets it determine whether it should update the color because it is local, or not update the color because it isn't local/the owner), than to demonstrate doing this kind of thing in HTML, which should imo be discouraged, and is just awkward and unclear and hard to follow and reason about.

See my previous comment.

> > Putting the old example https://glitch.com/edit/#!/naf-nametags in the repo and updating it would be great.
> 
> Agreed, and I'm glad to do this. I'd perhaps like to go as far as updating all of our demos to use nametags, possibly instead of colors, though, instead of having one example for nametags. I personally think nametags are a better choice than colors, because it's obvious that the nametag value comes locally from the user themselves, and that other nametags from other users come from networked (and not local) sources (which as far as I can tell, is supposed to be the point of including the color-randomizing concept in the demos).

I'll keep the colors for the simple reason it's easy to test multi users with several tabs in the same browser.
If you have only nametag, probably store the nametag in localStorage to not reenter it each time, you will end up with the same nametag for the avatars on all your tabs, unless you modify the nametag in a form that is accessible while in the room.

kylebakerio:
> From your long paragraph, I don't know if you understood it or not. :)
Which components and data of those components that will be synchronized is defined by the networked schema, not the template. You can have some entities with specific components in the template that are not synchronized, if there are not defined in the networked schema.

This exactly matches my understanding. You seem to be missing my point. (I'm also not sure if it's because you're not a native english speaker, but your use of a smiley here feels pretty condescending...?)

There are three declarations of the colored body sphere in our example:

1. The schema
2. The template
3. The HTML declaration in the body

Currently, for the color sphere:
1. You declare in `schema` what should be synced
2. You declare in `template` a consistent pattern that can be generated by all clients by NAF
3. You declare in `HTML` that for your own local instance, don't actually generate a copy, but do generate a copy on all other users' environments.
4. You also declare in `HTML` a duplicate of a part of what would have been generated according to the template.

This creates the problem for the user: how do I set/update the syned color of my sphere?

The solution we give in our demo is... declare an invisible duplicate of what is delcared in the template, in the HTML itself, and modify the synced property there. As long as you give it the right class and put it in the matching order of where the template _would_ have put it, it'll magically be synced as expected.

This is _terrible_ dev UI. I'm sure most users who look at this source struggle to understand what is going on here. I know I did when I first looked at it. From this example, it isn't clear that you generaly _don't need to declare a duplicate of what is in the template within the HTML_--just a reference to it is far more typical (I've never used this pattern from the demos in my code). It also takes some thought to figure out how/why that sphere is kept in sync--one would expect that only the entities generated by NAF according to the template are kept in sync, not that some other matching entity the user creates happens to be correctly picked up and synced as well. This behavior isn't explained anywhere, either--it's just given as a code example, and is in the only example we have.

Personally, I find it unnecessary to even make an API to handle this through the HTML directly, and would prefer we show how to do this in JS. I don't recall/know off the top of my head if NAF already creates the given elements even when `attachTemplateToLocal` is false and just makes them invisible, or if it doesn't add them at all, but either

1. It could add them and just add `visible=false`, not sync the visible property, and thus allow the user to modify them as needed
2. NAF could expose its own API for updating synced elements/components via JS that performs this element creation behind the scenes if needed.

> How do you detect who the owner is?
networked-audio-source.js does this:

```js
    NAF.utils.getNetworkedEntity(this.el).then((networkedEl) => {
      const ownerId = networkedEl.components.networked.data.owner;

      if (ownerId) {
        NAF.connection.adapter.getMediaStream(ownerId)
          .then(this._setMediaStream)
          .catch((e) => naf.log.error(`Error getting media stream for ${ownerId}`, e));
      } else {
        // Correctly configured local entity, perhaps do something here for enabling debug audio loopback
      }
    });
```

When I wrote the networked-video-source component, I had dug into this code deeply and knew the behavior off the top of my head, but I haven't touched NAF specific code in several months now. But just quickly glancing here, it seems that NAF doesn't add `networked.data.owner` to local components of networked elements. (Off the top of my head I would assume you match the owner to your own NAF ID, but it seems this is how the library currently does it.)

> I'll keep the colors for the simple reason it's easy to test multi users with several tabs in the same browser.
If you have only nametag, probably store the nametag in localStorage to not reenter it each time, you will end up with the same nametag for the avatars on all your tabs, unless you modify the nametag in a form that is accessible while in the room.

You don't need to store nametag in localstorage for the demos, and you can generate default random nametags that the user can override if desired. You can show a `prompt` that requests a username (and is prefilled with a default) prior to network connection. In fact, making it even more clear, we could add an overlay saying what user's perspective you have--currently, you can't tell what color you are unless you look at yourself from another user's screen. Nametags would be more clear than this, especially with a simple overlay showing one's own nametag value.

vincentfretin:
> This exactly matches my understanding. You seem to be missing my point. (I'm also not sure if it's because you're not a native english speaker, but your use of a smiley here feels pretty condescending...?)

Not at all my intend. I usually put smiley to indicate I mean no harm in the thing I just said, because sometimes I think my reply could be interpreted the wrong way by the other person. I had no idea it could be interpreted even worse with a smiley.
I genuinely didn't know if you understood or not how the things was working, maybe because of the user story way you phrased it and ending your sentence with a question mark.
From your long reply, it makes no doubt now you understand the issue well. And I'm looking forward to have a lively conversation with you on the subject. I'll reply to other parts later.

vincentfretin:
I agree with you with the poor developer experience on this subject. I also struggled to understand it by just reading the example.

> I don't recall/know off the top of my head if NAF already creates the given elements even when attachTemplateToLocal is false and just makes them invisible, or if it doesn't add them at all

It doesn't create the entities when attachTemplateToLocal is false.

> but NAF could either:
> 1.    add them and just add visible=false, not sync the visible property, and thus allow the user to modify them as needed (i.e., automatically do what the user is doing here behind the scenes).
> 2.    expose its own API for updating synced elements/components via JS that performs this element creation behind the scenes if needed (so, same as 1, but on demand--adds more complexity, but allows higher efficiency by only creating entities if needed)
> 3.    more efficient version of 2, where it doesn't actually do the creation, but just propogates the necessary changes. this is more code complexity, though, as we now have to keep track of syncing these virtual entities that are not attached to local. that said, this isn't crazy and should be pretty reasonable.

For 1, As a developer I don't expect those invisible entities to be created if attachTemplateToLocal is false. Depending on your template, this could have side-effects like downloading some resources like textures you didn't expect, and those may never be used for rendering. This really depends on your app use case though.
For 2, I understand it, but I'm not sure how this API could look like. Do you have an API design in mind?
For 3, if I understand this correctly, for example creating the sphere for the head with material component (because we sync .head material.color) but not the two spheres of the eyes based on what is synchronized in the networked schema? This seems a good idea, without the drawbacks of 1. I'm not sure how complex this implementation would be.

> You don't need to store nametag in localstorage for the demos, and you can generate default random nametags that the user can override if desired. You can show a prompt that requests a username (and is prefilled with a default) prior to network connection. In fact, making it even more clear, we could add an overlay saying what user's perspective you have--currently, you can't tell what color you are unless you look at yourself from another user's screen. Nametags would be more clear than this, especially with a simple overlay showing one's own nametag value.

I talked about localStorage because I have the `connectOnLoad:true` in mind and also as a user I don't want to reenter my name every time, but yes for demos we can simply generate random nametags and allow to change it when in the room via an overlay.
If you change to `connectOnLoad:false`, you can prompt for a nametag and color before joining the room.

kylebakerio:
> Not at all my intend. I usually put smiley to indicate I mean no harm in the thing I just said, because sometimes I think my reply could be interpreted the wrong way by the other person. I had no idea it could be interpreted even worse with a smiley. I genuinely didn't know if you understood or not how the things was working, maybe because of the user story way you phrased it and ending your sentence with a question mark. From your long reply, it makes no doubt now you understand the issue well. And I'm looking forward to have a lively conversation with you on the subject. I'll reply to other parts later.

Thanks for the clarification. I actually went and looked it up after this, and it's actually a common misunderstanding: https://www.bustle.com/life/what-does-the-smiley-face-emoji-happy-or-passive-aggressive

(That article makes it sound like it's mostly very young people, but I'm nearly 32, and my girlfriend--who is 28, and not a native english speaker--also felt the same way.)

For me, saying something like "I don't know if you understand it" and then putting a smiley face feels passive aggressive, but I understand that's not how you meant it. Glad I asked and we cleared that up.

kylebakerio:
> It doesn't create the entities when attachTemplateToLocal is false.

I thought so, thanks for clarifying.

> For 1, As a developer I don't expect those invisible entities to be created if attachTemplateToLocal is false. Depending on your template, this could have side-effects like downloading some resources like textures you didn't expect, and those may never be used for rendering. This really depends on your app use case though.

Yeah, this could incur some level of performance cost in theory in a few ways, but I think the effect is pretty minor. As long as `visible=false`, I don't think it will incur rendering burden, which is the most important thing, I'd think. It does create the best developer experience, though, imo.

Nothing special is needed. We could rename `attachTemplateToLocal` to something else to be more clear--we could also just add this as an alternate mode and leave both, depending on what is needed by the dev. `attachTemplateToLocal=false` for when the dev wants to make sure the template is completely not present, and something like `hideLocalTemplate=true` as an alternative when they want it to be there so they can propagate networked changes using the standard A-Frame component apis while still hiding it locally. 

> For 2, I understand it, but I'm not sure how this API could look like. Do you have an API design in mind?

Nothing that I have in mind is truly elegant--I really appreciate how seamlessly NAF works so that nothing like this is necessary--but an example might be something like 

```js
NAF.setAttribute('#element-id', 'component', 'property', 'value')
```

The idea is that when you use this, if you are using it on something with `attachTemplateToLocal=true` it does nothing special, but if it's with `attachTemplateToLocal=false` and the piece of the template is missing locally, then it would either:

- silently add in the necessary component, and then continue in the future by updating invisible, newly created component
- as per (3), just propoagate the change to peers without adding the missing template piece

> I talked about localStorage because I have the connectOnLoad:true in mind and also as a user I don't want to reenter my name every time, but yes for demos we can simply generate random nametags and allow to change it when in the room via an overlay.
If you change to connectOnLoad:false, you can prompt for a nametag and color before joining the room.

yeah, exactly. I also generally just don't add `networked-scene` until I'm ready and make a wrapper component to handle that logic, but same idea.

vincentfretin:
> > How do you detect who the owner is?
> 
> networked-audio-source.js does this:
> 
> ```js
>     NAF.utils.getNetworkedEntity(this.el).then((networkedEl) => {
>       const ownerId = networkedEl.components.networked.data.owner;
> 
>       if (ownerId) {
>         // not owned by user
>       } else {
>         // owned by user
>       }
>     });
> ```
> 
> When I wrote the networked-video-source component, I had dug into this code deeply and knew the behavior off the top of my head, but I haven't touched NAF specific code in several months now. But just quickly glancing here, it seems that NAF doesn't add `networked.data.owner` to local components of networked elements. (Off the top of my head I would assume you match the owner to your own NAF ID, but it seems this is how the library currently does it.)

I found it weird that owner is empty here for your own local components, so I dug a little bit.
Your code above was inspired by
https://github.com/networked-aframe/networked-aframe/blob/51d08ba6ca21b660a16df4f70b2cb91cb620fb4b/src/components/networked-audio-source.js#L22-L33
where it checks if owner is empty. For this particular component, it's used generally on the `avatar-template` and is created locally only if `networked="template:#avatar-template;attachTemplateToLocal:true;"`
`networked.data.owner` will indeed be empty initially before we are connected, but it is not empty anymore once you're connected. Once connected, the owner is set here https://github.com/networked-aframe/networked-aframe/blob/ac93c62a05b31f5c1747e0969904a04b4a7fbc48/src/components/networked.js#L267-L270
So this hack works only if you use the condition in `init`, it won't work in `update`, so be careful.
If you want a condition in `update` to check if it's the owner, you may want to check `!networked.data.owner || networked.data.owner === NAF.clientId`.

kylebakerio:
> So this hack works only if you use the condition in init, it won't work in update, so be careful.
If you want a condition in update to check if it's the owner, you may want to check !networked.data.owner || networked.data.owner === NAF.clientId.

Good catch! That's more like what I expected. That makes sense, and yes, that was me quickly checking how networked-audio-source.js does it because I just needed a quick reference for how it can be done, and that seemed like it should be a good example since it was in the repo. I was really surprised, and `networked.data.owner === NAF.clientId` is exactly what I expected it to be.

> Thanks for the link about the smiley happy or passive-aggressive, this is interesting. This is indeed a common misunderstanding it seems. I'll think twice before adding a smiley now. I'm using https://www.dictionary.com/e/emoji/grinning-face-with-sweat-emoji/ from time to time in chat, wow this can have a lot of meanings. And https://www.dictionary.com/e/emoji/grinning-face-with-smiling-eyes-emoji/ was really bad when the other person received a completely different smiley design.

That's funny. The 'smiling with sweat emoji' has _always_ been associated to me with its appearance in anime, which is mentioned further down on the link you gave when it comments here:

> You will also see this emoji accompanying posts about anime because, according to Winnie Liu of the Factorialist, “the giant sweat drop above the brow has traditionally appeared in Japanese animation…to indicate awkwardness, frustration, embarrassment, and sentiments like What the heck?”  

I just _always_ interpret it this way, not when it has anything to do with anime. I also haven't watched that much anime in my life--it's just the only place I had seen gigantic sweat drops on a smiling face, so that's just obviously what I interpreted it as. Even as a kid watching pokemon, for example, I was seeing images like this:

name

> https://www.dictionary.com/e/emoji/grinning-face-with-smiling-eyes-emoji/

This one was interesting, though I've never really used that one.

> For the point 2, I agree. For points 1 and 3, follow-up discussion on the associated issues.

will follow up and return here afterwards.

kylebakerio:
update: so, dug into our 'determine if an el is owner by local user or not` question, and it seems it _can_ work on update. With a little more experimentation, I've come up with this function:

```js
      function ownedByLocalUser(el) {
        return new Promise((resolve, reject) => {
          NAF.utils.getNetworkedEntity(el).then((networkedEl) => {
            const ownerId = networkedEl.components.networked.data.owner;

            if (ownerId || ownerId !== NAF.clientId) {
              resolve(false)
              // not owned by user
            } else {
              resolve(true)
              // owned by user
            }
          });
        })
      }
```
But I have to say... If this is really necessary, I feel like we need to update NAF so that it isn't.

kylebakerio:
Alright, newest version of the draft features:

- **updated color syncing**, as discussed in relevant issue. uses `visible="false"` to achieve the desired effect instead of `attachTemplateToLocal:false;`
- **have live-synced, auto-generated, user-settable, local-overlay nametags**
- **features new draft of onConnect concept**, as discussed in that issue

https://glitch.com/edit/#!/8-2021-naf?path=examples%2Ftracked-controllers.html%3A97%3A0

kylebakerio:
Hm. We should still make one nested component synced somewhere, though. it's good to have a simple example of how that works. That said, I almost never use this feature myself--might make sense to just feature that in a separate demo.

vincentfretin:
`NAF.clientId` is only set when successfully connected, The `clientId` may come from the server, so we don't know it before we're connected. I think your `NAF.utils.safelyGetOwner` will continue to return empty string in `init` when not currently connected.
Anyway we don't currently need this for our use case, we just check if some element like `.head` is there to differentiate between local or remote if `attachTemplateToLocal:false`. If `attachTemplateToLocal:true` we don't even need to differentiate between local or remote.

kylebakerio:
> `NAF.clientId` is only set when successfully connected, The `clientId` may come from the server, so we don't know it before we're connected. I think your `NAF.utils.safelyGetOwner` will continue to return empty string in `init` when not currently connected. Anyway we don't currently need this for our use case, we just check if some element like `.head` is there to differentiate between local or remote if `attachTemplateToLocal:false`. If `attachTemplateToLocal:true` we don't even need to differentiate between local or remote.

That's right. I tried to break it up into two smaller functions, but the only reason I needed `safelyGetOwner` was to support `ownedByLocalUser`. That particular break in `safelyGetOwner` isn't a problem for `ownedByLocalUser`, but it would undermine it if used directly for another purpose--anyways, that's all a side thing.

As you say, we can just dodge the issue entirely. I still think we should make this easier--functionality like this is pretty basic, and is more necessary when you're writing general purpose components, like `networked-audio-source`--but I agree it isn't really necessary for this demo.

Updated now to:

```js
      AFRAME.registerComponent('player-info', {
        schema: {
          name: { type: 'string', default: "user-" + Math.round(Math.random()*10000) },
          color: { type: 'string', default: '#' + new THREE.Color( Math.random(), Math.random(), Math.random() ).getHexString() },
        },

        init: async function() {
          this.head = this.el.querySelector('.head');
          this.nametag = this.el.querySelector('.nametag');
          this.ownedByLocalUser = this.el.id === "local-head";
          if (this.ownedByLocalUser) {
            this.nametagInput = document.getElementById("username-overlay");
            this.nametagInput.value = this.data.name;
            this.nametagInput.oninput = () => {
              this.el.setAttribute('player-info', 'name', this.nametagInput.value || " ");
            };
          }
        },

        update: async function(newArgs) {
          if (this.head) this.head.setAttribute('material', 'color', this.data.color);
          if (this.nametag) this.nametag.setAttribute('value',this.data.name)
        }
      });
```

kylebakerio:
Is there anything else? Are we settled on this at this stage? If so, I think the only thing left would be updates to onConnect, but will need to do that pull request first before changing that, and so I'd recommend we just:

- accept these changes now
- later I'll update onConnect, and with that pull request, will update this demo along with the documentation

vincentfretin:
>     * maybe looking at updating the schema API, that's pretty rough right now and needs to be fixed (are you working on this, or interested in doing it? If not, it's bothering me and I'd like to see it improve, and could at least do a high level wrapper around this method)

I'm not working on it, and fixing it is not currently on my agenda. If you find a way to fix the issue in a PR from the comments I wrote in #267, we'll discuss it then.

Issue #269, opened 2021-04-06
Multiple video streams for the easyrtc adapter


vincentfretin:
We added camera streaming in networted-aframe 0.8.0 with #261 and a fix just after in 0.8.2 https://github.com/networked-aframe/networked-aframe/commit/2ae6b3b8eda74091452ff7bdeb84ab4f2d890a4e

It would be great if we could be able to easily add screen sharing or an arbitrary video stream with camera on, so two video tracks or more.
On Open-EasyRTC side you can register a new stream using the `easyrtc.register3rdPartyLocalMediaStream` API.
```
navigator.mediaDevices.getDisplayMedia({video: true}).then(function (stream) {
    easyrtc.register3rdPartyLocalMediaStream(stream, 'screen:1');

    // Then to add to existing connection
    easyrtc.addStreamToCall(easyrtcId, 'screen:1', function (caller, streamName) {

    });
});
```

The janus adapter has an API `setLocalMediaStream(stream)` (stream includes zero or one audio track and zero or one video track) that you can call several times. It will replace or disable the audio or video track on the RTCPeerConnection, but in janus-plugin-sfu it's limited to one audio track and one video track on the RTCPeerConnection.

We could expand on this API to allow multiple video tracks.
- `addLocalMediaStream(stream, streamName)`
- `removeLocalMediaStream(stream, streamName)`
- `getMediaStream(clientId, streamName)` (currently streamName can only be audio or video, can be now any name?)

There are lots of things to considerate here, the renegotiation of the SDP offer/answer when adding and removing a video track, what to do at the aframe component level when a video stream is ended, and maybe more I didn't think about.

vincentfretin:
@keianhzo @netpro2k for the dialog adapter and @mqp for the janus adapter. If we can agree on the same API for multiple tracks, that would be great. If you have some insights that I didn't think about, please share.

In the janus adapter there is 
- `enableMicrophone(enabled)`

We added it in the easyrtc adapter with:
- `enableCamera(enabled)`

I see in the [dialog adapter](https://github.com/mozilla/hubs/blob/master/src/naf-dialog-adapter.js) you have:
- `async enabledMic(track)`
- `async enableCamera(track)`
- `async disableCamera()`
- `async enableShare(track)`
- `async disableShare()`
- `enableMicrophone(enabled)`

jimver04:
I have done some checks here and I think this is an open-easyrtc issue. I have opened an issue there:

[...]

PR #294, opened 2021-09-13
Multiple video streams for the easyrtc adapter #269

hthetiot:
- add `addLocalMediaStream`, `removeLocalMediaStream` and update `getMediaStream`
- add streamName option to video and audio components
- No breaking change for previous API usage
- Add sample with screen stream on top of default video stream (camera).

Screenshot 2021-09-13 at 18 15 54


hthetiot:
One thing that can be interesting is to add a stream option in the networked-scene shema to use stream if provided instead of create one. alternatively calling addLocalMediaStream at any point before or after connection will works out of the box in this PR.

src/adapters/EasyRtcAdapter.js
```diff
@@ -173,58 +183,114 @@ class EasyRtcAdapter extends NoOpAdapter {
...
+
+    // Add local stream to existing connections
+    Object.keys(this.remoteClients).forEach((clientId) => {
+      if (easyrtc.getConnectStatus(clientId) !== easyrtc.NOT_CONNECTED) {
+        easyrtc.addStreamToCall(clientId, streamName);
+      }
+    });
+  }
+
+  removeLocalMediaStream(stream, streamName) {
```
vincentfretin, line 288: I would have simplified it to just `removeLocalMediaStream(streamName)` Wouldn't it be better? This way you don't need to keep the stream object in a variable after you call `addLocalMediaStrem` to call `removeLocalMediaStream` later.

vincentfretin:
Thanks. I'll do a proper review and some testing in the coming days.
There is a typo in the filename basic-mutli-stream.html => multi
Could you please add the new API `addLocalMediaStream`/`removeLocalMediaStream` and `streamName` field for the `networked-video-source` to the documentation in the README?

> One thing that can be interesting is to add a stream option in the networked-scene shema to use stream if provided instead of create one

I didn't understand. The schema allows only primitive types I think, no? You can't set a stream on an entity field.
Why do you say "instead of creating one"? Which one are you talking about?

hthetiot:
https://github.com/networked-aframe/networked-aframe/pull/294#discussion_r708195913

I agree, i did not want to change the API you instructed because I thought it had to march what other adapter does. I can change not problem.

hthetiot:
> There is a typo in the filename basic-mutli-stream.html => multi
Thx will fix.

> Could you please add the new API addLocalMediaStream/removeLocalMediaStream and streamName field for the networked-video-source to the documentation in the README?


Ok will do.

hthetiot:
> I didn't understand. The schema allows only primitive types I think, no? You can't set a stream on an entity field.

I see, Idid not know that.

> Why do you say "instead of creating one"? Which one are you talking about?

Yes I thought user may want pass custom stream from schema i did not realised the primitive types limitation.
In any case user can use addLocalMediaStream at any point in time in the life cycle.

vincentfretin:
> [#294 (comment)](https://github.com/networked-aframe/networked-aframe/pull/294#discussion_r708195913)
> 
> I agree, i did not want to change the API you instructed because I thought it had to match what other adapter does. I can change not problem.

Ah no, no other adapter have this API. I didn't see it was me I specified this in the issue. I just didn't think much when I wrote the issue :)

hthetiot:
Just need to add the doc and should be good, will do later today may be.

examples/basic-multi-streams.html
```diff
@@ -0,0 +1,121 @@
...
+
+        <img id="grid" src="https://img.gs/bbdkhfbzkk/stretch/https://i.imgur.com/25P1geh.png" crossorigin="anonymous">
+        <img id="sky" src="https://i.imgur.com/WqlqEkq.jpg" crossorigin="anonymous" />
+
+        <!-- Templates -->
+
+        <!-- Avatar -->
+        <template id="avatar-template">
+          <a-entity class="avatar">
+             <a-plane  color="#FFF" width="4" height="3" position="0 .6 0"  material="side: front" networked-video-source></a-plane>
```
vincentfretin, line 35: Can you keep the `side: back` for the webcam as well like we had for basic-video.html?
```
             <a-plane  color="#FFF" width="4" height="3" position="0 .6 0"  material="side: back" networked-video-source></a-plane>
```
It's hard to see if there is another person in the room otherwise when testing. On one of my test, I thought the second participant wasn't there but it's just I was behind him. :)

src/adapters/EasyRtcAdapter.js
```diff
@@ -173,58 +183,113 @@ class EasyRtcAdapter extends NoOpAdapter {
...
+      // Add mediaStreams audio streamName alias if does not exist yet
+      const audioTracks = stream.getAudioTracks();
+      if (!clientMediaStreams.has('audio') && audioTracks.length > 0) {
+        // Safari doesn't like it when you use single a mixed media stream where one of the tracks is inactive, so we
+        // split the tracks into two streams.
+        const audioStream = new MediaStream();
+        try {
+          audioTracks.forEach(track => audioStream.addTrack(track));
+          clientMediaStreams.set('audio', audioStream)
+        } catch(e) {
```
vincentfretin, line 242: I think this is wrong here. You should set the audioStream outside the `if (this.pendingMediaRequests.has(clientId))` like we had before. If I'm not mistaken, `setMediaStream` can be called before any `getMediaStream` (when the entity is not yet created, `networked-video-source` component didn't call `getMediaStream` yet), so in this case the `if (this.pendingMediaRequests.has(clientId))` will be false and will never set the audio and video.
This is probably the reason for the regression I have on my iPad.

src/adapters/EasyRtcAdapter.js
```diff
@@ -246,14 +311,17 @@ class EasyRtcAdapter extends NoOpAdapter {
...
+    this.easyrtc.setOnStreamClosed(function(clientId, stream, streamName) {
+      delete this.mediaStreams[clientId][streamName];
     });
 
     if (that.easyrtc.audioEnabled || that.easyrtc.videoEnabled) {
-      this.easyrtc.initMediaSource(
+      navigator.mediaDevices.getUserMedia({
+        video: that.easyrtc.videoEnabled,
+        audio: that.easyrtc.audioEnabled
+      }).then(
```
vincentfretin, line 322: This is what `this.easyrtc.initMediaSource` was doing?
Maybe not in the PR, we should be able to specify options here 
`audio: {echoCancellation: true, noiseSuppression: true, autoGainControl: true}` instead of `audio: true`
and for example
`video: {mediaSource: "camera", width: { max: 1280, ideal: 640 }, height: { ideal: 360 }, frameRate: 30}` instead of `video: true`

vincentfretin (changes requested):
Sorry for delay, I took the time only now to reconfigure https on the nodejs process, reconfigure my PC and test with two devices on the same network.

On the basic-video.html example I have a regression with these changes, on master this works well between Chrome Ubuntu and iPad 14.8 Safari
With the changes, the second participant only see a white screen for the other participant. No error in chrome console, on Safari I don't know I don't have easy access to the logs.

For the basic-multi-streams.html example, I think I succeeded only once on the iPad to see the two streams of the other participant. But after that I never saw again the second participant, on the PC on Chrome console I don't even see "Creating remote entity" so there is probably a js error on Safari, maybe randomly. I didn't dig further.

Do you reproduce the regression on basic-video.html?

I don't have much devices with me to test, I have currently only my PC on Ubuntu and iPad. My Android device refuse to connect on my hostname.local domain I don't know why.

I'll retest next weekend once the basic-video.html is working again properly for me probably after a fix in setMediaStream if you have time to look into it, see my comment above.
I'll also update to iPadOS 15 in the meantime, and try to use vorlon server to access the Safari logs.

hthetiot:
> Sorry for delay, I took the time only now to reconfigure https on the nodejs process, reconfigure my PC and test with two devices on the same network.

No problem, we all have job on the side or open source 🐱 , i will check when I have a moment myself. Thank you for the review and testing.

For https you should check ngrok for free https tunnel to localhost (`npm i -g ngrok` and `ngrok http 8080`).

vincentfretin:
I updated my iPad to iPadOS 15.0.1, same issue with the regression.

With Android Chrome, I tested via the ip https://192.168.1.14:8080 instead of https://mymachine.local:8080 and accepted the certificate. Same issue with white screen on Android Chrome. It works on master.
I couldn't install ngrok, the server that host the archive currently times out.

vincentfretin (approved):
I merge this and create another PR with the changes in my branch.

src/adapters/EasyRtcAdapter.js
vincentfretin: I created the issue #300 to track this.

examples/basic-multi-streams.html
vincentfretin: The camera was actually really rendered behind the avatar. This is why I had a hard time testing this. :)
I fixed it in https://github.com/networked-aframe/networked-aframe/commit/38bd9773e7ce973091ffcd3850eb699a22ecc0da

examples/basic-multi-streams.html
```diff
@@ -0,0 +1,121 @@
...
+            screenBtnEle.textContent = 'Share screen';
+          } else {
+            navigator.mediaDevices.getDisplayMedia().then((stream) => {
+              NAF.connection.adapter.addLocalMediaStream(stream, "screen");
+              screenEnabled = true;
+              screenBtnEle.textContent = 'Stop Screen';
+            });
+          }
+        });
+      }
```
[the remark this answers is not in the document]
vincentfretin: Issue created #301 for this.

vincentfretin:
The example is now on https://naf-examples.glitch.me/basic-multi-streams.html

PR #298, opened 2021-10-31
Additional changes for Multiple video streams for the easyrtc adapter

vincentfretin:
Additional changes for #294


vincentfretin:
The example is now on https://naf-examples.glitch.me/basic-multi-streams.html

Issue #287, opened 2021-07-19
Where is easyrtc.js?


ckasarda:
In the example code, there is a reference being made to a file that no longer exists, called easyrtc.js. Does anyone know where this went or what replaces it? Thank you!

The line causes errors is this: `<script src="/easyrtc/easyrtc.js"></script>`

lazarmitic:
Download that file from examples and host it on your server.

ckasarda:
Awesome thanks it works now!

vincentfretin:
It's weird you need to copy it, did you properly install dependencies with `npm install`?. See also my reply #286.

vincentfretin:
I don't reproduce the issue with easyrtc not found.
If you follow the getting started instructions and copied the server and examples directories, ran `npm install` and `npm start`, the only instruction that may be missing to run an example is to edit `examples/shooter.html`
to replace
```
<script src="/dist/networked-aframe.js"></script>
```
by
```
<script src="https://unpkg.com/networked-aframe@^0.8.0/dist/networked-aframe.min.js"></script>
```
and http://localhost:8080/shooter-2.html is properly working.

vincentfretin:
Cool, I added a note in https://github.com/networked-aframe/networked-aframe/commit/7a34d3ca964d6c0d3b533dd152a0fc16a0876585
We actually had this issue since I removed this little hack https://github.com/networked-aframe/networked-aframe/commit/11c4909ff925ea8dd906e1a00f8a339e1231c846 
I prefer to document it instead of having this hack used in the wild, which did a 404 Not Found on `/dist/networked-aframe.js` and then downloaded the latest version, version that could have been different from the one specified for the server part in `package.json`.

PR #306, opened 2021-11-22
In the instructions, you forgot to copy the dist file

jimver04:
Hi Vincent, I am starting a new local naf . Following the instructions, my basic example .html does not work because dist folder is not copied to installation folder. I have put the robocopy and cp command to fix it in this pull request.

### DIFF
--- a/docs/getting-started-local.md
+++ b/docs/getting-started-local.md
@@ -60,10 +60,12 @@ Let's copy the example server that comes with networked-aframe into our project.
 # MacOS & Linux
 cp -r ./node_modules/networked-aframe/server/ ./server/
 cp -r ./node_modules/networked-aframe/examples/ ./examples/
+cp -r ./node_modules/networked-aframe/dist/ ./examples/dist/ /e
 
 # Windows
 robocopy .\node_modules\networked-aframe\server\ .\server\ /e
 robocopy .\node_modules\networked-aframe\examples\ .\examples\ /e
+robocopy .\node_modules\networked-aframe\dist\ .\examples\dist\ /e
 ```
 
 You'll now see another new folder: `server/`. Inside it you'll see `easyrtc-server.js` which is the server code.


vincentfretin:
Hi, my opinion concerning the getting started documentation is to explain to the developer to use a released version of networded-aframe, not a copy of the dist folder that will be eventually outdated.
I did a change explaining this just below your changes https://github.com/networked-aframe/networked-aframe/commit/7a34d3ca964d6c0d3b533dd152a0fc16a0876585
See also my comment https://github.com/networked-aframe/networked-aframe/issues/287#issuecomment-945066845
What do you think? Can you explain this better? Move some paragraphs or change the explanation maybe?
It may be tedious to modify the index.html of each example if you want to test all of them locally?
Should we revert back to using the hack we had before to use a released version at unpkg.com if the dist folder is not found?
This:
```
<script src="/dist/networked-aframe.js"></script>
<script>window.NAF || document.write('<script src="https://unpkg.com/networked-aframe@^0.8.0/dist/networked-aframe.min.js">\x3C/script>')</script>
```
I personally don't like it because you get those lines in the wild, it doesn't look like a good practice to me.
To avoid a 404 Not Found error for /dist/networked-aframe.js in a production environment you should really modify those two scripts tags and replace it by a single script tag:
```
<script src="https://unpkg.com/networked-aframe@0.8.2/dist/networked-aframe.min.js"></script>
```
So reverting to this and explain what I just said in the documentation?
Let me know what you folks think about it.

jimver04:
It is annoying to have examples that actually not work because a line is commented a priori. So,

either
1) the line about naf.js should be uncommented in the examples
or
2) the dist folder should be copied in the instructions for installation

I prefer the second as most of the people are interested on the latest version. In such a case, please remove /e from the line that regards cp dist command in the instructions for local installation (my fault, I was working in windows, but now that I have tested in Linux the /e  is not a parameter of cp but only for robocopy).

Best,
Dimitrios

vincentfretin:
Got it, I did the changes in https://github.com/networked-aframe/networked-aframe/commit/98d635f2ed9c55167b67fa0beea326f95200e513
Thanks for the feedback.

Issue #296, opened 2021-10-28
New Demo of Networked-Aframe with MediaPipe Self-Segmentation


jimver04:
There is no place for demos so I am adding it here !

We have made a stage for live concerts using Networked-Aframe and MediaPipe. 

Video

[https://www.youtube.com/watch?v=2d8P70C7N_4](https://www.youtube.com/watch?v=2d8P70C7N_4 )

![image](https://user-images.githubusercontent.com/1898502/139278992-f90bb199-7620-48f0-bf7a-2fbadef3b0ea.png)
In the comments of the video you can find the Demo link for live testing. Currently in Beta. Since MediaPipe is sometimes glitchy, you may need to refresh Firefox or Chrome. Use a pc, laptop, or tablet. On mobiles (Chrome only) there might be issues.

vincentfretin:
@jimver04 The multi streams example is now merged
https://github.com/networked-aframe/networked-aframe/blob/master/examples/basic-multi-streams.html
and live here
https://naf-examples.glitch.me/basic-multi-streams.html

If you can adapt your demo to use the new API and give feedback on it that would be awesome.
You should be able to use `networked-scene="audio:true;video:false"` to not have the camera enabled, and use 
`NAF.connection.adapter.addLocalMediaStream(canvasStream, "mediapipedCamera");`
and rewrite your `networked-video-source-mediapiped` to use latest changes of `networked-video-source` and use it like this `networked-video-source-mediapiped="streamName:mediapipedCamera"`
Be aware of known issues #299 though. I don't know it you have currently those issues in your demo.

vincentfretin:
https://webrtchacks.com/how-to-make-virtual-backgrounds-transparent-in-webrtc/ is a similar implementation than yours.

jimver04:
@vincentfretin Yes we have copied it from the same library of Mediapipe  👍 
 I have tried to make video avatars to co-exist with 3D avatars but there is a problem with streams. We have an evaluation at our project this week and I do not have time. I am also preparing an authoring tool for artists to make their own environment as a plugin for WordPress using Three.js.  I will definitely come back on this on the next week. Best, D. 

https://www.youtube.com/embed/WataNoHgjlo
![image](https://user-images.githubusercontent.com/1898502/139870755-3d89f10c-2bc3-442b-b2c5-c3122521cc87.png)

jimver04:
Sorry for not replying earlier but I was engaged in another project in UE4. 

The steps are as follows:
[...]

Issue #291, opened 2021-08-31
Cheap virtual production with NAF and Mediapipe


jimver04:
Well my target is to make a multiplaying 3D environment where the persons are represented by cubes that have as textures their selfie without the background. It is a kind of cheap virtual production without chroma key background. The background is removed with MediaPipe Selfie-Segmentation wasm javascript code. The issue is that instead of having the other player texture on the cube (P1 should see P2, and P2 should see P1, each one sees his selfie. This means that P1 sees P1 and P2 sees P2 which is bad.

The full description of the issue can be found in 
https://stackoverflow.com/questions/68994847/using-self-segmentation-javascript-of-mediapipe-to-pass-user-selfie-as-a-texture

I have used : this.video.srcObject = canvasElement.captureStream(30); to capture the canvas video and add it as a texture to the cube. Is this correct ?

vincentfretin:
I see you took the networked-video-source component that you modified.
In `networked-video-source` you have this line:
```
 this.video.srcObject = newStream;
```
You replaced it by:
```
this.video.srcObject = canvasElement.captureStream(30);
```

I don't see how your canvasElement draw the video track from newStream here.
It seems that selfie_segmentation take your webcam by default? You need to configure selfie_segmentation to use the video track from newStream. I'm not familiar with the lib, so I can't help you with it.

But be careful in my opinion you should not do the post processing on the receiving end, because here you're still sending your webcam with your background to all the other users, then you want to do the post processing when receiving each participant video? This won't be great for performance if you have more that 2 users. And this gives a false sense of security for the user. You still can show the webcam from another user with their background with a few lines of javascript in the console. :)

What you should normally is get the webcam, do the post-processing, then set the resulting canvas as a video track to the peer connection. I don't know how to do that with easyrtc, but I did a similar example with the janus adapter where I stream the scene canvas https://github.com/networked-aframe/naf-janus-adapter/blob/master/examples/audio-with-camera.html#L187-L189

vincentfretin:
You may also be interested by the camera filters experiment that Hubs did here https://github.com/mozilla/hubs/pull/4352
that uses https://github.com/tensorflow/tfjs-models/tree/master/body-pix
Hubs is using their mediasoup adapter, but you can achieve the same with the janus adapter that also implements the `NAF.connection.adapter.setLocalMediaStream` API.

Good luck!

jimver04:
Hi Vincent,

[...]

PR #310, opened 2022-01-25
Multistreams with Green Screen Effect

jimver04:
Hi Vincent,

I have used the MultiStreams example to make an example for green screen effect.

Contribution:

examples/advanced-video-green-screen.html
src/components/networked-video-source-green-screen.js

Real Demo:
https://vrodos-multiplaying.iti.gr/advanced-video-green-screen.html

Tests:
Firefox (not yet tested with Chrome)

Methodology Explanation:

When you send stream: I use a canvas to process the camera stream and send the one without green pixels.
When you receive a stream (for all received streams): I use a custom shader to remove all green pixels.

Video:
A video about a previous version here:
https://www.youtube.com/watch?v=WKKZUz_iKqY

More collaboration:
Let's make a meeting for demonstration. For the time being, I have dat.gui handles for only green pixels (g-r>thres). In the todo is to add also other colors as well as YUV encoding rather than RGB which is more useful in compositing.

Optimization:
The original camera stream is disabled to save some bandwidth.  I have used 800x600 to keep the bandwidth around 1Mbps for two persons. I guess that 0.5 Mbps is added per person. 

Best,
Dimitrios


docs/getting-started-local.md
```diff
@@ -371,3 +371,33 @@ I would love love love you to send me cool examples you've made, and I'm looking
...
+
+You want the second field namely ProcessID which happens for me to be 3766 but for you will be another number
+
+
+**2. Kill**
+
+**sudo kill -9 3766**
+
+
+
```
vincentfretin, line 403: On Linux, a process that runs in a terminal in foreground can be stopped with Ctrl+C.
Is this not working for you?

server/easyrtc-server.js
```diff
@@ -8,8 +8,8 @@ const easyrtc = require("open-easyrtc");      // EasyRTC external module
 // Set process name
 process.title = "networked-aframe-server";
 
-// Get port or default to 8080
-const port = process.env.PORT || 8080;
+// Get port or default to 5832
+const port = process.env.PORT || 5832;
```
vincentfretin, line 12: Please don't change the default port in the PR.

src/components/networked-video-source-green-screen.js
```diff
@@ -0,0 +1,164 @@
...
+
+  update: function () {
+
+    var data = this.data;  // Component property values.
+    var el = this.el;  // Reference to the component's entity.
+
+    if (data.event) {
+      // This will log the `message` when the entity emits the `event`.
+      el.addEventListener(data.event, function () {
+      });
```
vincentfretin, line 22: dead code?

src/components/networked-video-source-green-screen.js
```diff
@@ -0,0 +1,164 @@
...
+  setupVideo: function() {
+    if (!this.video) {
+      const video = document.createElement('video');
+      video.setAttribute('autoplay', true);
+      video.setAttribute('playsinline', true);
+      video.setAttribute('muted', true);
+      this.video = video;
+    }
+  }
+});
```
vincentfretin, line 164: The idea I had was to make changes in the networked-video-source component, maybe adding a parameter or two to have the option of using a custom shader instead of writing a specific networked-video-source-green-screen, duplicating lots of the code.

vincentfretin (commented):
Quick review with some comments. I didn't look in details to the example yet. I'm going to run them now.

server/easyrtc-server.js
vincentfretin: If you need to change the port, you can do this in bash in your terminal:

```
export PORT=5832
npm start
```

and even put this line in your `~/.bashrc`
so next time you just use `npm start` and this will use the redefined port.

examples/advanced-video-green-screen.html
```diff
@@ -0,0 +1,227 @@
...
+  });
+
+  // Called by Networked-Aframe when connected to server
+  function onConnect () {
+    console.log("onConnect", new Date());
+    setTimeout(function(){
+      // This is the processed stream that we want to send
+      // We do not want the default stream camera (takes bandwidth)
+      NAF.connection.adapter.enableCamera(false);
+    },5000);
```
vincentfretin, line 203: Did you try setting `video:false` on networked-scene like I said in https://github.com/networked-aframe/networked-aframe/issues/296#issuecomment-955667234 ?
As far I understand, having `video:true` implies negotiating a video track with the other participants, and setting `enableCamera(false)` afterwards will just make the track inactive. Setting `video:false` will not add the default camera video track from the RTCPeerConnection at all.
I don't have two PC to test that on my local machine. I usually test with both my PC and iPad, but on the iPad the page freeze.

examples/index.html
```diff
@@ -81,9 +81,35 @@ <h1>Networked-Aframe Examples</h1>
...
+    <h2>Advanced Video Examples</h2>
+
+
+
+    <div>
+      <table>
+        <tr>
+          <td style="width:200px">
+          </td>
+          <td style="width:600px">
```
vincentfretin, line 95: Why this empty cell that make an indentation?

If you want to separate the basic examples from advanced examples, I'm fine with it, but you should probably rewrite the h1 at the top to be just "Networked-Aframe", add a h2 Examples, and here the h2 Advanced Video Examples so it's at the same level I think. And use the same syntax as the Examples section instead of using a table.

examples/index.html
```diff
@@ -81,9 +81,35 @@ <h1>Networked-Aframe Examples</h1>
...
+
+
+
+
+
+    <br />
+    <br />
+    <h3>
+    <p>Want to help make better examples? That would be awesome! <a href="https://github.com/networked-aframe/networked-aframe/blob/master/CONTRIBUTING.md#join-the-community-on-slack" target="_blank">Contact us on Slack</a>.</p>
+    </h3>
```
vincentfretin, line 112: A p tag inside h3 is not valid. https://stackoverflow.com/questions/19779519/is-it-valid-to-have-paragraph-elements-inside-of-a-heading-tag-in-html5-p-insid

vincentfretin:
I like the dat gui to tweak the value for the custom shader, that's great for a demo. This was working properly for me on Firefox and Chrome, but not Safari. But I couldn't test the networked part of it though.
Only GreenThresholdIn is really used, GreenThresholdOut is not used AFAICT.

I don't know much how aframe handle additional material, but I think you can create only once the custom shader, and reuse the same for every networked-video-source component instance, but using specific uniforms params?

I think for this PR to be merged and the new example to be maintained in the long term is just to keep your example advanced-video-green-screen.html and try to have only one networked-video-source component with necessary changes instead of duplicating it.

I see you committed the file basic-video-mediapiped.html. I think this was your previous example that was not using the new api. You probably want to remove the file from the PR. 

If possible I think you should put an updated mediapipe example on glitch that is using the new networked-aframe api+this PR with changes in networked-video-source supporting the custom shader. The goal in the end is to have your mediapipe example running with minimal lines of code on a released version of networked-aframe, and not a custom build.
And we could put a link to your glitch in the examples.

src/components/networked-video-source-green-screen.js
```diff
@@ -0,0 +1,164 @@
...
+        this.materialIncoming.fragmentShader = `
+                           varying vec2 vUv;
+                           uniform sampler2D uMap;
+                           uniform float GreenThresholdIn;
+                           
+                           void main() {
+                                vec2 uv = vUv;
+                                vec4 tex1 = texture2D(uMap, uv * 1.0);
+                                 if (tex1.g - tex1.r > GreenThresholdIn)
+                                    gl_FragColor = vec4(0,0,0,0);
```
vincentfretin, line 113: Did you try discard like I said in https://github.com/networked-aframe/networked-aframe/issues/291#issuecomment-939309492 ?

vincentfretin:
Look at `registerShader` https://aframe.io/docs/1.2.0/components/material.html#registershader
and look a the `Programs` count of the stats component
https://aframe.io/docs/1.2.0/components/stats.html
I think with your current code, this Programs counter will keep increasing for each new participant (I can't confirm it right now), but if you properly reuse the shader you will only have one I think.

docs/getting-started-local.md
jimver04: The process might be initialized with a cron job or some other way. We usually put an automatic way in case for some reason the server restarts. So , in order to kill it you must search it.

server/easyrtc-server.js
jimver04: ok thanks, I didn't know that way !

src/components/networked-video-source-green-screen.js
jimver04: oops. it was from debugging. Yes delete it.

src/components/networked-video-source-green-screen.js
jimver04: My mainly concern was not to break the existing component and have something for you to merge without conflict resolution. I can merge them if it is an issue.

examples/advanced-video-green-screen.html
jimver04: I will try that also. I totally forgot it as it has been a long time ago.

examples/index.html
jimver04: I think we should add a representative screenshot for each example. It is easier thus to select. I will fix it. I am preparing an example.

src/components/networked-video-source-green-screen.js
vincentfretin: It doesn't make much sense to me to have two different components that are almost identical in the NAF library. It should be a properly registered shader and an option to the existing `networked-video-source` to use a custom shader. Of course the default options should have no change to the existing examples, so still using the default shader.
It depends on the level of contribution you are willing to take. If you decide to keep it separated, in this case you shouldn't include it in `src/components` but keep it only for your example, so in `examples/js` directory.

vincentfretin: Actually you shouldn't need to change networked-video-source at all. You should be able to register a new "green-screen" shader with `registerShader`, modify your shader to use uniform `map` instead of `uMap`. And use in your entity `networked-video-source="streamName: screen" material="shader:green-screen;GreenThresholdIn:0.02"`

docs/getting-started-local.md
vincentfretin: Ok. Generally pm2 is used to deploy a nodejs app, to start it as a service. Last time I deployed easyrtc-server.js on a test server I read https://www.digitalocean.com/community/tutorials/how-to-set-up-a-node-js-application-for-production-on-ubuntu-16-04 and used this:
```
$ sudo npm install -g pm2
$ pm2 start easyrtc-server.js
                          Runtime Edition

        PM2 is a Production Process Manager for Node.js applications
                     with a built-in Load Balancer.

                Start and Daemonize any application:
                $ pm2 start app.js

                Load Balance 4 instances of api.js:
                $ pm2 start api.js -i 4

                Monitor in production:
                $ pm2 monitor

                Make pm2 auto-boot at server restart:
                $ pm2 startup

                To go further checkout:
                http://pm2.io/


                        -------------

[PM2] Spawning PM2 daemon with pm2_home=/home/vincentfretin/.pm2
[PM2] PM2 Successfully daemonized
[PM2] Starting /home/vincentfretin/vr/easyrtc-server.js in fork_mode (1 instance)
[PM2] Done.
┌────────────────┬────┬──────┬────────┬───┬─────┬───────────┐
│ Name           │ id │ mode │ status │ ↺ │ cpu │ memory    │
├────────────────┼────┼──────┼────────┼───┼─────┼───────────┤
│ easyrtc-server │ 0  │ fork │ online │ 0 │ 0%  │ 29.6 MB   │
└────────────────┴────┴──────┴────────┴───┴─────┴───────────┘
 Use `pm2 show <id|name>` to get more details about an app


vincentfretin@server1:~/vr$ pm2 startup
[PM2] Init System found: systemd
[PM2] To setup the Startup Script, copy/paste the following command:
sudo env PATH=$PATH:/usr/bin /usr/lib/node_modules/pm2/bin/pm2 startup systemd -u vincentfretin --hp /home/vincentfretin

[PM2] Freeze a process list on reboot via:
$ pm2 save

[PM2] Remove init script via:
$ pm2 unstartup systemd

Show info about the app:

    pm2 show easyrtc-server

Show logs:

    tail -f /home/vincentfretin/.pm2/logs/easyrtc-server-out.log
```

Feel free to modify your doc to include that.

jimver04:
As regards the procedure of replacing pixels, it is as follows. Get my video stream and replace "green" pixels with flat RGBA = (0,1,0,1) pixels (flat green). The amount of "green" of the background is configured with GreenThresholdOut. This is due to the fact that there are always shadows on your background as no green is absolute flat green. For this procedure I have used an html canvas and traverse it pixel by pixel.

When you receive all the "Greenish" video streams from other clients, you should be able to remove the green pixels. This can be done by simply saying if RGBA == (0,1,0,1)  then put transparent (0,0,0,0). But for experimentation reasons I have added also a GreenThresholdIn  which is a threshold on the amount of green for the received streams (for all the clients that you receive a video stream). It is mainly for debugging. For this method I have used the shader language. 

I am planning to do something for the white channel also as many have white background. It would be lossy however, as eye bulbs and reflections are white. A custom color picker would be also useful. Thank god there is dat.gui for that. 

As regards Mediapipe for automatic selfie segmentation with AI, I will re-visit it when there will be an improvement by Google. The last version 0.6 was very problematic. Only 50% of the times worked and the error codes were not illuminating the problem.

vincentfretin:
Thank you for the explanation, it's clearer now what your loop does, replacing the greenish pixels with pure green pixels.
I completely skipped your loop when I read your code, I see now that `GreenThresholdOut` is used there.
You should put this explanation in comments just above your `loop()`.

And as I understand it with your mediapipe version you did a similar thing, filling the detected background by pure green (`canvasCtx.fillStyle = "#00FF00FF";`). Cool.

Maybe I'm asking something obvious, is there a reason to compare the difference between red and green, instead of just checking green above some threshold?

For your loop, wouldn't it be better to use `setInterval` instead of `setTimeout`? I don't know if it makes a difference for performance.
https://developer.mozilla.org/en-US/docs/Web/API/setInterval

vincentfretin:
I reply to myself here, `setInterval` would potentially execute again the function even if the previous call wasn't finished I think.

vincentfretin:
For anyone interested in playing with the new mediapipe selfie segmentation:
https://twitter.com/TensorFlow/status/1488214410346872840?t=bEJ0d-1WnvU4lsLxhR-z5w
If you don't have a green background :)

src/components/networked-video-source-green-screen.js
jimver04: I have selected the last case, however,

*How do I update the map with the video texture ?*  I am completely stack with the shader below. My map is always null. I have not much experience with shaders and there only a few examples in Aframe.

```
<script>

    let vertexShader1 = `
                           varying vec2 vUv;

                          void main() {
                              vec4 worldPosition = modelViewMatrix * vec4( position, 1.0 );
                              vec3 vWorldPosition = worldPosition.xyz;
                              vUv = uv;
                              gl_Position = projectionMatrix * modelViewMatrix * vec4( position, 1.0 );
                          }
                        `;

    let fragmentShader1 = `
                             varying vec2 vUv;
                             uniform sampler2D map;
                             uniform float gthreshold;

                             void main() {
                                  vec2 uv = vUv;
                                  vec4 tex1 = texture2D(map, uv * 1.0);
                                  if (tex1.g - tex1.r > gthreshold)
                                      gl_FragColor = vec4(0,0,0,0);
                                  else
                                      gl_FragColor = vec4(tex1.r,tex1.g,tex1.b,1.0);
                              }`;


    AFRAME.registerShader('green-screen', {

        schema: {
            map: {type: 'map', is: 'uniform'},
            gthreshold : {type: 'float', value: 0.02}
        },

        uniforms: {
            // the texture value (once the texture source is loaded, update)
            map: { type: 't', value: null },
            gthreshold:  {type: 'f', value: 0.02}
        },

        init: function(data) {

            this.material = new THREE.ShaderMaterial({
                uniforms: this.uniforms,
                vertexShader: this.vertexShader,
                fragmentShader: this.fragmentShader
            });
        },

        update: function (data) {
            AFRAME.utils.material.updateMap(this, data);
            this.uniforms.map.value = this.material.map;
        },

        vertexShader: vertexShader1,
        fragmentShader: fragmentShader1
    });

</script>
```

examples/advanced-video-green-screen.html
jimver04: Yes you are right, it works with video:false

examples/index.html
jimver04: I will keep the old style

examples/index.html
jimver04: I didn't have any decent screenshot. I will make a better scene in the future

src/components/networked-video-source-green-screen.js
jimver04: Today it worked with 
discard;

jimver04:
As regards "Programs" in stats, you have right, it increases incrementally when each player enters. However, I do not know how to use shaders for video streams. There is no example anywhere.

vincentfretin:
I'll take a look for the shader code this weekend.

jimver04:
I have found how to make a green-screen shader for videos in AFrame without the need of a canvas and without the need to go back to Three.js.

Look at two examples at my aframe 1.3.0 fork 
Shader based:  https://github.com/jimver04/aframe/tree/master/examples/test/shader-texture-video
Canvas based:  https://github.com/jimver04/aframe/tree/master/examples/test/canvas-texture-video

I don't know which is faster. My GPU is 3080 and doesn't have any problems showing both at 60fps. Is there any method to check it?  Stats are showing about the same.

image

I will see to fix NAF now. 
Best,
D.

jimver04:
The video dom example is not working properly in NAF as it shows always the same video for all players. I guess I have to use the embedded map of the shader 

```
/* global AFRAME */
/**
 * The Green screen shader
 */
AFRAME.registerShader('green-screen-shader', {

    schema: {
        greenThreshold:{type: 'float', is: 'uniform', default:0.08}
    },

    vertexShader: `varying vec2 vUv;

                          void main() {
                              vec4 worldPosition = modelViewMatrix * vec4( position, 1.0 );
                              vec3 vWorldPosition = worldPosition.xyz;
                              vUv = uv;
                              gl_Position = projectionMatrix * modelViewMatrix * vec4( position, 1.0 );
                          }`,

    fragmentShader: `varying vec2 vUv;
                     uniform sampler2D map;
                     uniform float greenThreshold;

                     void main() {
                                  vec2 uv = vUv;
                                  vec4 tex1 = texture2D(map, uv * 1.0);
                                   if (tex1.g - tex1.r > greenThreshold)
                                      gl_FragColor = vec4(0,0,0,0);
                                   else
                                      gl_FragColor = vec4(tex1.r,tex1.g,tex1.b,1.0);
                              }`

});
```
However the map corresponds to the whole level map but not the player map texture:

image

jimver04:
By Searching aframe.js, the only map declared for flat shader (the shader of videos) is src. I have replaced map with src but I get the same map, i.e. the whole level map.

jimver04:
Finally I have done a simple solution. It is very robust in comparison to my other methods.
- No secondary stream
- No canvas for personal stream processing before send
- Just apply a green effect shader for all incoming video streams. 
Although it does not give the ability for each user to configure its booth setup, it is much better as regards speed and accessibility.
More advanced-examples will follow in the future. 
1) use canvas per user to configure personal stream locally before sending
2) Check new versions of MediaPipe (cross-fingers with WebAssembly)

I think we can not avoid Three.js as Aframe does not apply a shader with registerShader when no dom element for texture is defined. It applies only globaly in the scene to the first texture it finds, e.g. to the skysphere in my case.

vincentfretin:
I didn't look at the `registerShader` with custom shader with the `map` I talked about yet, but it's on my todo list.
I was busy with the other issues and doing the releases in the last weeks. ;-)

What are you up to in the latest commits?
FYI you're working on the master branch in your fork, so all the commits ends up in this PR.
For futures PR you may want to create a branch instead.

jimver04:
Hi Vincent, 

my crazy idea is to do remote virtual productions for shooting movies whilst the actors are at their homes. It is very risky but I have managed to attract the interest of the local community of actors and musicians through a demo, as well as persuading the project managers. Actors will participate through their smartphones from home with a green screen background* and with a low rendering needs page. They only have to press 3 keys that will transfer them in their acting position, so as to avoid 3D navigational stress.  I will be in the same room as a director with my RTX3080 card rendering the full scene:

image
Director High-end PC

then I stream back the overall scene with the "Screen sharing" feature that you have recently added in order for the actors to have a feedback:

image
Actor Smartphone


Most of the developments are on the director page. I have followed your advice to avoid using canvas on the sender side and use only shaders on the receiver side. I have changed the RGB algorithm to YUV space with double thresholds which is more commonly used in chroma key removal (see https://www.shadertoy.com/view/MlVXWD). I have also added these thresholds on the Director page together with the choice to control the chroma key (it can be also blue or any color). I have added dat.gui interfaces per actor so that each actor can have separate chroma key and different thresholds. I have also added the parameters to change the width and the height of each actor plane so that I can normalize the size of the actor in the scene. 

I have followed your advice, to avoid doing processing on a canvas on the sender side and the code became more simple. I am working only with shaders and it works well. I am using a modified version of your adaptor, and I have delete mine.


Of course the nice Ship 3D model is excluded from all the commits as I have bought it from a well-known site. 
The water is done with A-water component (https://github.com/Dante83/a-water) which is open source.

image


To summarize, ignore this pull request. I will prepare another branch depending on what you want from all these above. 
Any comments ? ideas ? My future plan is to add some movement for the actors' planes to give them the ability to walk in the scene. It is not possible for them to touch their smartphone so it has to be done on the Director side. 

Best,
Dimitrios


*Web AI for autoremoval (Mediapipe) is unstable and computational expensive for smartphones, plus it has artifacts of the background. We will share to actors some low cost "Creator studios" consisting of a green cloth sheet and a smartphone stand ~ 70euros.

vincentfretin:
Nice project! Thanks for keeping us posted about the things you're trying, this is interesting.
I'll try to tinker a bit the custom shader on my side and let you know.

Issue #280, opened 2021-05-08
Need newer buffered-interpolation version with newer threejs (r128)


vincentfretin:
The documentation is wrong, we don't use [aframe-lerp-component](https://github.com/haydenjameslee/aframe-lerp-component) since may 2018 https://github.com/networked-aframe/networked-aframe/commit/e15f34250c2c28e00ade2c57cf94f9f0043f0c1b. We use [buffered-interpolation](https://github.com/InfiniteLee/buffered-interpolation) to update position, rotation, scale.
=> Update the README.md and CONTRIBUTING.md

vincentfretin:
@netpro2k merged a similar PR https://github.com/InfiniteLee/buffered-interpolation/pull/7
Does somebody other than @InfiniteLee have the right to publish a new version on npmjs.com?

vincentfretin:
Note that the slerpQuaternions change is not compatible with aframe 1.2.0 / three r125, so I won't include this change in a new networked-aframe version until aframe updates threejs to at least r128.

vincentfretin:
If you have your own aframe build with threejs r128, you can create your own networked-aframe build with in `package.json`:
```
"buffered-interpolation": "github:Infinitelee/buffered-interpolation"
```

vincentfretin:
Folks are still using aframe 1.0.4 or 1.1.0 with latest networked-aframe. So if aframe updates threejs and we update networked-aframe to use the latest version of buffered-interpolation, it would mean networked-aframe won't support old aframe versions anymore. :(

Maybe we can find a way to support both deprecated and new api in buffered-interpolation in our own fork? We will think about it again when aframe update to recent threejs.

vincentfretin:
I don't think adding a condition to check THREE version in the [`InterpolationBuffer.slerp`](https://github.com/InfiniteLee/buffered-interpolation/blob/1940a8202019f0c5ecc0e25ae43c292b8cc449d7/index.js#L61-L63) method would be a good idea for perf reason, because this is executed on each tick.
Maybe define `this.slerp` inside the InterpolationBuffer class constructor to define a different slerp method based on THREE version? I didn't test but maybe something like that could work?

```
class InterpolationBuffer {
  constructor(mode = MODE_LERP, bufferTime = 0.15) {
    if (parseInt(THREE.REVISION, 10) < 128) { // or lower 126 or 127?
      this.slerp = function(target, r1, r2, alpha) {
[...]

PR #313, opened 2022-02-06
Update to aframe 1.3.0

vincentfretin:
Update to aframe 1.3.0 and latest buffer-interpolation version #280

Issues you may encounter when upgrading your project to aframe 1.3.0:
- If your model appears darker than with aframe 1.2.0, try adding `renderer="colorManagement: true"` to `<a-scene>`
- If you use the environment component and it appears broken, be sure to update it to version 1.3.1.


vincentfretin:
I see some warnings
`THREE.Quaternion: Static .slerp() has been deprecated. Use qm.slerpQuaternions( qa, qb, t ) instead.`
in the shooter and google blocks examples. Not sure where those come from. Someone has an idea?

vincentfretin:
I see the stacktrace of the warning now, on Chrome. I was testing with Firefox last time I think and didn't see it.
I have the warning when there is a second participant. With `npm run dev`

```
THREE.Quaternion: Static .slerp() has been deprecated. Use qm.slerpQuaternions( qa, qb, t ) instead.
slerp @ three.js:2618
slerp @ buffered-interpolation.js:106
update @ buffered-interpolation.js:213
tick @ networked.js:288
``

wait..., executing `npm start` fixed the issue. `npm start` is doing a build a copy it to `examples/dist/`
but `npm run dev` seemed to use an old build instead of rebuilding it in memory. If I remove the files `examples/dist/` and run `npm run dev`, this is properly working, the build is done in memory.
Ok, there is probably something I don't understand about webpack, or a misconfiguration? Bah nevermind. Let's merge this.

PR #318, opened 2022-02-26
update engines to node>=16 npm>=7

vincentfretin:
Update engines to node>=16 npm>=7 so glitch selects node v16.12.0 npm 7.20.6 environment instead of node 10.

The new package-lock.json lockfile v2 format requires npm >= 7.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,8 +83,8 @@
     "webvr"
   ],
   "engines": {
-    "node": ">= 8.10.0",
-    "npm": ">= 4.0.5"
+    "node": ">=16",
+    "npm": ">=7"
   },
   "babel": {
     "presets": [


vincentfretin:
I updated https://glitch.com/edit/#!/naf-project
from 
```
"engines": {
  "node": "8.10.0"
},
```
to
```
"engines": {
  "node": ">=16"
},
```

to force using glitch environment
```
$ node -v
v16.12.0
```
instead of
```
$ node -v
v8.17.0
```


See supported nodejs versions https://nodejs.org/en/about/releases/

https://glitch.happyfox.com/kb/article/59-can-i-change-the-version-of-node-js-my-project-uses/

https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines

vincentfretin:
I reverted this change in 1aaa0e2afc34f98e9298de8a6d7705a549d0f63d
This is good change for a project and to deploy on glitch, but here networked-aframe is also a library, this would mean you will be required to use node 16 yourself. Not a good thing to do this change in a patch release that's for sure, and I'm planning to do a 0.9.1 release now.

vincentfretin:
I added a note for myself to edit the engine on glitch after I resync the repo there 570700357a691957757bcda64c0dc445506d9ea9

PR #326, opened 2022-03-13
Add a persistent sphere example

### LINKED ISSUE #238: Add an example of a shared sphere
Add an example in the repo of a shared sphere where each participant can click on it to change its color.
see discussion https://aframevr.slack.com/archives/C65MC3NSF/p1611344560006300
with the glitch https://glitch.com/edit/#!/naf-material-color that use a sphere in the html with `persistent:true`

vincentfretin:
This closes #238

Dynamically spawned persistent spheres is another use case entirely. I'll work on an advanced example.

### DIFF
--- a/examples/basic-persistent.html
+++ b/examples/basic-persistent.html
@@ -1,0 +1,187 @@
+<html>
+  <head>
+    <meta charset="utf-8">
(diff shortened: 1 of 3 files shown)


vincentfretin:
You can test it at https://naf-examples.glitch.me/basic-persistent.html

Issue #265, opened 2021-03-24
Add more doc about persistent entities


vincentfretin:
See #238 to add an example that use a persistent entity with `persistent:true`.

About `persistent:true` I made some comments with my findings here https://aframevr.slack.com/archives/C65MC3NSF/p1616577974040300
that I paraphrase here:

With `persistent:true`, each participant is responsible of creating the entity, it is only synced. Generally you have an entity in the html like the example in #238 or you create the entities dynamically before you enter the room, this is what Hubs does.

If you're using `connectOnLoad:true`, you may receive an update for a persistent entity that is not created yet, the data is saved for later, see https://github.com/networked-aframe/networked-aframe/blob/393ec7b8ffe22ff20e4b2f84fe6f32a7c23f7c06/src/NetworkEntities.js#L78-L81
In an old version of Hubs, they used that with the `applyPersistentFirstSync` api to update the newly created entity with the data previously received. I searched in the history, here is an example https://github.com/mozilla/hubs/commit/f2d779dab7c252ded4ca5f6aac448022b070b40d
Today, Hubs doesn't use this API anymore, they maintain their own list of persistent entities received through Phoenix socket, and use their own api `applyPersistentSync` https://github.com/mozilla/hubs/search?q=applyPersistentSync

There is also an option `NAF.options.firstSyncSource` that is not documented https://github.com/networked-aframe/networked-aframe/blob/393ec7b8ffe22ff20e4b2f84fe6f32a7c23f7c06/src/NetworkEntities.js#L75-L77
to ignore messages for persistent entities coming from a source that we don't want (naf adapter), this is used by hubs to only take care of the persistent message via the Phoenix socket https://github.com/mozilla/hubs/blob/614818eb2a1c3b23b49700a2023709a86532e283/src/hub.js#L225

The persistent attribute was added to the table https://github.com/networked-aframe/networked-aframe#example-attachtemplatetolocalfalse in the documentation (that's the only doc!) by Hayden in commit https://github.com/networked-aframe/networked-aframe/commit/64a852fb2e46d529c1c449c06d68a23517e142ef just after he merged the changes from Mozilla's fork that contains this PR https://github.com/MozillaReality/networked-aframe/pull/33
see the related PR on Hubs too https://github.com/mozilla/hubs/pull/1556

PR #327, opened 2022-03-26
Add a spawned persistent spheres (peer to peer) example

vincentfretin:
The example is similar to #326 but here with spawned spheres.
I created a `spawner-persistent` component and `persistent-p2p` that handle the logic of keeping the persistent entities peer to peer. The spheres are kept in the room if there is at least one participant remaining in the room.

This is related to #265, it's the first time we use the `applyPersistentFirstSync` api in an example.

Next example would be persisting it server side.


vincentfretin:
Example live at https://naf-examples.glitch.me/persistent-peer-to-peer.html

Issue #324, opened 2022-03-12
Make the easyrtc adapter unreliable transport configurable


vincentfretin:
From the discussion in #139
make the easyrtc adapter unreliable transport configurable (used for `sendData` and `broadcastData`) , so you can use websocket (like the wseasyrtc adapter) or a custom function to use a websocket from another backend instead of webrtc data channels.
With websocket transport you can do whatever you want on the server side to persist spawned entities.

In the janus adapter you can specify for reliableTransport and unreliableTransport "datachannel", "websocket" or a function to use another server. Example in the janus `sendData` code:
https://github.com/networked-aframe/naf-janus-adapter/blob/7840de8f2dbfd978136c7e1844d4677fbfd208fe/src/index.js#L1037-L1053

Issue #139, opened 2018-07-21
Is there a best practice to persist changes to scene on the server?


hilderonny:
The minecraft demo is impressive for multiplayer games. But how would you implement a server feature to store the changes made by the users in a session on the server and loading them when a new user connects?

Currently the changes on the scene are sent via datachannels directly to the other users, am I right? So there is no way for the server to protocol those changes because the server does not see what happens in the scene, correct?

How should one implement such a persistence feature? Should the server "join" the room as a client to get the changes via a data channel? Or should one client forward the changes to the server via an easyrtc event message?

btw.: Great work. Looking forward to use it in different projects.

haydenjameslee:
Yes that's correct re: datachannels.

There are many ways to do state persistence. You could use a service like firebase, or write your own server to manage it, or use the same signalling server. There's certainly no best practice.

I would start with firebase personally.

haydenjameslee:
I think it'd be really cool to make a persistence framework on top of NAF (if anybody is looking for project ideas).

HeadClot:
I would recommend looking into [apatite.js](https://github.com/apatitejs/apatite) opposed to reinventing the wheel so to speak. I would try to get that to work with networked A-Frame. It already works with Nodejs and a bunch of various databases already.

vincentfretin:
In the janus adapter you can specify for reliableTransport and unreliableTransport "datachannel", "websocket" or a function to use another server. It would be great to have this option on the easyrtc adapter too.

vincentfretin:
There was a discussion about that on this twitter thread https://twitter.com/vincentfretin/status/1502322662571487236 where I wrote some comments.

vincentfretin:
With websocket transport you can do whatever you want on the server side to handle `um` (list of entity updates), `u` (entity update), `r` (entity removed) messages with `persistent: true` (`isFirstSync: true` when spawning, then `false` if the entity is moved) to persist the spawned entities in the room. Each entity has a random `networkId`.

vincentfretin:
You'll probably want to keep a Map of persistent entities with networkId key in memory, with a dirty flag when you modify it and throttle saving that Map in a postgres json field. This is what I would do.

To load this json and create the entities when you join a room, see #265

vincentfretin:
To persist in postgres, we can use the open source https://supabase.com that has a similar api than firebase, probably with a docker-compose file to setup the needed containers. But having a full example with it may be out of scope for this repo.

PR #328, opened 2022-03-26
Add a spawned persistent spheres (server side) example

vincentfretin:
#327 was using a peer to peer persistence.
In this example, we save the persistent entities server side. This addresses #139

- [x] keep in memory server side the persistent entities (using the NAF `u` persistent firstSync messages may not be the right way though, this requires to have unreliable transport on the websocket, so currently with wseasyrtc, see #324 to use websocket for easyrtc adapter)
- [ ] broadcast persistentEntityCreated message to all participants from the server when a participant create a persistent entity
- [ ] send `persistentEntities` message from the server when a participant join the room
- [ ] `persistent-server` component that do the necessary things with the `persistentEntityCreated` and `persistentEntities` messages coming from the server
- [ ] use NAF option `firstSyncSource` to allow firstSync entities from the server only, see #265
- [ ] persist entities in a file on the filesystem (persisting in postgres via [superbase](https://supabase.com/) out of scope for this repo)

### DIFF
--- a/server/easyrtc-server.js
+++ b/server/easyrtc-server.js
@@ -67,6 +67,88 @@ easyrtc.events.on("easyrtcAuth", (socket, easyrtcid, msg, socketCallback, callba
     });
 });
 
+const ROOM_OBJECTS = {
+  _dirty: false
+};
+
+const NAF_MSG_TYPES = ["u", "um", "r"];
+easyrtc.events.on("easyrtcMsg", (connectionObj, msg, socketCallback, next) => {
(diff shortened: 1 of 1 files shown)


vincentfretin:
To save on the server persistent entities, saving all components updates is probably wrong in some cases.
If you spawn for example a video, you want to save the position, rotation, scale and url of the video, but not the `video.currentTime` on the server. But you still want the video current time to be synced between participants.
So you want to save only if isFirstSync is true.

PR #330, opened 2022-03-27
handle update of positional param in networked-audio-source

### LINKED ISSUE #245: switch from non positional to positional in networked-audio-source
If I use for example in my template `networked-audio-source="positional:false"` and later want to switch to positional audio with
`setAttribute('networked-audio-source', 'positional:true;rolloffFactor:2.0;distanceModel:exponential')` I get an error.
I did get around it with this code in my app but this is quite ugly, this needs to be properly handled by the networked-audio-source update function.
```
    // change to positional audio
    const players = document.querySelectorAll('[player-info]');
    for (let i = 0; i < players.length; i++) {
      if (players[i].id !== 'cameraRig') {
        const nas = players[i].querySelector('.camera').components['networked-audio-source'];
        const stream = nas.stream;
        nas.remove();
        nas.sound = null;
        nas.stream = null;
        players[i].querySelector('.camera').setAttribute('networked-audio-source', 'positional:true;rolloffFactor:2.0;distanceModel:exponential');
        nas._setMediaStream(stream);
      }
    }
```

This should really be:
```
    // change to positional audio
    const players = document.querySelectorAll('[player-info]');
    for (let i = 0; i < players.length; i++) {
      if (players[i].id !== 'cameraRig') {
        players[i].querySelector('.camera').setAttribute('networked-audio-source', 'positional:true;rolloffFactor:2.0;distanceModel:exponential');
      }
    }
```

vincentfretin:
This closes #245

To test it, open http://localhost:8080/basic-audio.html and connect two users
and move away from the other avatar so you almost don't hear them.

in the console:
```
const players = document.querySelectorAll('.avatar');
// switch to non positional, you should now hear the other avatar
for (let i = 0; i < players.length; i++) { players[i].setAttribute('networked-audio-source', 'positional:false'); }
// switch to positional, the sound of the other avatar should be attenuated
for (let i = 0; i < players.length; i++) { players[i].setAttribute('networked-audio-source', 'positional:true'); }
```

### DIFF
--- a/src/components/networked-audio-source.js
+++ b/src/components/networked-audio-source.js
@@ -1,5 +1,5 @@
 /* global AFRAME, NAF, THREE */
-var naf = require('../NafIndex');
+const naf = require('../NafIndex');
 
 AFRAME.registerComponent('networked-audio-source', {
   schema: {
@@ -15,17 +15,18 @@ AFRAME.registerComponent('networked-audio-source', {
   },
 
   init: function () {
-    this.listener = null;
     this.stream = null;
+    this.sound = null;
+    this.audioEl = null;
 
-    this._setMediaStream = this._setMediaStream.bind(this);
+    this.setupSound = this.setupSound.bind(this);
 
     NAF.utils.getNetworkedEntity(this.el).then((networkedEl) => {
       const ownerId = networkedEl.components.networked.data.owner;
 
       if (ownerId) {
         NAF.connection.adapter.getMediaStream(ownerId, this.data.streamName)
-          .then(this._setMediaStream)
+          .then(this.setupSound)
           .catch((e) => naf.log.error(`Error getting media stream for ${ownerId}`, e));
       } else {
         // Correctly configured local entity, perhaps do something here for enabling debug audio loopback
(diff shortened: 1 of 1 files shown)


vincentfretin:
I'm introducing a new option `createSoundOnInit` (default true) and `sound-source-removed` event to properly disconnect the soundSource of an audio analyzer for example.
This refactoring would allow for example to use `createSoundOnInit:false` and handling creating or destroying the audio externally.
A use case I have in mind is hundreds of avatars, and allow only audio of avatars near you. (There is a lot more to do to support this use case, in particular changes in the easyrtc or janus adapter), but this is a first step.

```
const players = getNearbyPlayers();
for (let i = 0; i < players.length; i++) {
  const nas = players[i].components['networked-audio-source'];
  if (!nas.sound) { nas.createSound(); }
}
```

if a player is far away, destroy the audio source
```
const nas = player.components['networked-audio-source'];
if (nas.sound) { nas.destroySound(); }
```

vincentfretin:
I didn't merge this yet because I'm working on an example so that I can test the proposed api further.

vincentfretin:
Wow time flies. I'll merge in this state, examples will maybe come later.

PR #331, opened 2022-03-30
Set the dirty flag when updating

takahirox:


vincentfretin:
No problem, I see you recreated it in the mozilla fork https://github.com/MozillaReality/networked-aframe/pull/47
The changes are probably also ok in the original repo, even if we don't have the matrix auto update false by default here.

vincentfretin:
Ah no ok `matrixNeedsUpdate` doesn't exist in Object3D, it's a specific flag that you use in hubs patches. So yeah we probably don't want the changes here. :)

PR #332, opened 2022-04-06
Update buffered interpolaation version

keianhzo:
There is a mismatch between the hubs client buffered-interpolation version:
https://github.com/mozilla/hubs/blob/a3c5b169031a5c24ab8d11355d72420667c296e5/package.json#L85
and the `networked-aframe` one:
https://github.com/MozillaReality/networked-aframe/blob/master/package.json#L29
That causes issues when updating the package.json in the hubs client as it reverts it to the `networked-aframe` version. This PR consolidates both.


keianhzo:
Wrong destination

Issue #347, opened 2022-07-05
remix demo is broken


kylebakerio:
https://glitch.com/edit/#!/remix/naf-examples

always returns an error, won't load, won't even let you see the code. :/

kylebakerio:
huh. Seems like it was likely some temporary downtime. working now…

kylebakerio:
https://naf-examples.glitch.me/tracked-controllers.html

kylebakerio:
this old copy is accessible, though I'm not sure what/how much has changed since then: https://glitch.com/edit/#!/8-2021-naf?path=examples%2Ftracked-controllers.html%3A97%3A0

kylebakerio:
this link does work: https://glitch.com/edit/#!/remix/naf-project

kylebakerio:
I contacted support and got some details:

> Good news: the app is running! You are very close to hitting your maximum disk space, and the npm install took 10 minutes. I think the first issue is that the ">= 16" node engine wasn't working and so the container was defaulting to Node 10 (we're going to look into why that string doesn't work). The other issue is that when npm installs take a long time, it can make it look like the editor is hanging because 10 minutes is a long time! Your project appears to have 2 webpack builds which could add to this time as well.

> I think this project like this would work best while boosted (which gives it more memory and disk space). I'll check on the project tomorrow to make sure it's running and let you know if we discover anything else.

(they did not update me the next day)

We should probably simplify this glitch demo.

vincentfretin:
I'm using glitch for hosting the examples, I don't think we should promote remixing naf-examples, this includes all the tests stack you don't need and is doing the naf build (twice?), that's why it takes a long time to start up. You should start from naf-project template if you're starting a new glitch, it's using a released version of naf.
FYI I wrote in https://github.com/networked-aframe/networked-aframe/blob/master/docs/release-instructions.md how I resync this glitch. Specifying 
```
"engines": {
    "node": ">=16"
  },
```
worked before when I tested it in #318 and I see it's working properly in https://glitch.com/edit/#!/naf-project
Oh, I see now in the downloaded glitch archive it's using 
```
"engines": {
    "node": "16"
  },
```
Did I make a mistake last time I edited the file? Probably.

For naf-examples, currently I have the message "This project has been suspended".
I'll look into it and send email to the support once I figured out a new build/start strategy for the examples.
[...]

PR #348, opened 2022-07-12
Don't build on start but only on install

vincentfretin:
See #347 for context.
I introduced some time ago (in https://github.com/networked-aframe/networked-aframe/commit/017d40cca82b0f7666327e0a14a8f49b3224b2ba and https://github.com/networked-aframe/networked-aframe/commit/a97b5e79dd30785557c2f33d184d3741f8315a14) running `npm run build` as part of `npm start` to be sure to have latest build when using `npm start` instead of `npm run dev`. In retrospect, I think this change was wrong, we shouldn't run a build step on `npm start` (for glitch) so I'm reverting this change.
`npm install` is executing `npm run prepare` that is executing `npm run dist` already. I added the `cp -r dist examples` in the prepare script instead. And I fixed the webpack-dev-server config to use the bundle in-memory and not the one from examples/dist when running `npm run dev` (no need to run `npm run dist` before running `npm run dev`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,4 +41,4 @@ jobs:
         run: npm run lint
 
       - name: Check Build
-        run: npm run build
+        run: npm run dist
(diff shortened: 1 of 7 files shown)


vincentfretin:
I added a simple `package-glitch.json` for glitch to avoid installing all the tests stack and doing a build. I overwrite the `package.json` file with this file when I resync manually glitch after a release.

PR #358, opened 2022-08-15
Update examples: add `controller` option

kylebakerio:
- added `controller` option! 
- also, support controller model (and button model) animations!
- more semantic use of NAF `createdByMe()` to remove need for class
- clean up `this.Y[this.Z]` pattern by more correctly utilizing A-Frame prototype pattern 
- support for live-switching between models
- support for many other controller models (not only vive, but also older models as well)

see live updated demo: https://naf-wip.glitch.me/tracked-controllers.html


kylebakerio:
@vincentfretin is it alright to merge in some updates to the component? Some significant code improvements--also added a cool 'controller' option that allows sync'd controllers to be used. Core existing functionality that was already present is unchanged.

src/components/networked-hand-controls.js
```diff
@@ -153,38 +106,269 @@ AFRAME.registerComponent('networked-hand-controls', {
...
+    if (oldData.handModelStyle &&
+        oldData.handModelStyle !== this.data.handModelStyle
+      ) {
+      // first, remove old model
+      this.el.removeObject3D(this.str.mesh);
+      ['gltf-model','obj-model'].forEach(modelComponent => {
+        if (this.el.components[modelComponent]) {
+          this.el.components[modelComponent].pause();
+          this.el.components[modelComponent].remove();
+          this.el.removeAttribute(modelComponent);
```
vincentfretin, line 120: Why do you call pause and remove yourself? It should already be done with the removeAttribute.

src/components/networked-hand-controls.js
```diff
@@ -153,38 +106,269 @@ AFRAME.registerComponent('networked-hand-controls', {
...
+
+      // this prevents injectTrackedControllers from running, which prevents the 
+      // tracked-controls-webxr component from being added, which is responsible for pose tracking
+      // (which we are disabling, since we want this controller to follow this NAF entity instead)
+      this.el.components[this.data.controllerComponent].checkIfControllerPresent = x => x;
+      this.el.components[this.data.controllerComponent].injectTrackedControls = x => x;
+      
+      // we load the model indicated by the remote user's headset
+      this.el.components[this.data.controllerComponent].loadModel({profiles:JSON.parse(this.data.webxrControllerProfiles)});      
+      this.el.play();
```
vincentfretin, line 329: Why do you need to call play yourself here?

src/components/networked-hand-controls.js
```diff
@@ -153,38 +106,269 @@ AFRAME.registerComponent('networked-hand-controls', {
...
+        // 25 seconds, don't want the possiblity of this running forever in the background
+        clearInterval(disableInterval);
+        try {
+          this.el.components[this.data.controllerComponent].remove();
+        } catch(e) {
+          console.error("error removing controller component", e);
+        }
+        throw new Error("never found tracked-controls-webxr, cancelling interval that was waiting to remove it and removing controller components")
+      }
+    },25)
```
vincentfretin, line 361: Can you please use the tick method instead of setInterval?

src/components/networked-hand-controls.js
kylebakerio: If you look at these controller components, some (all?) of them don't have remove() specified, while the pause() functionality does things that should (also) be done in remove, like remove event listeners. I call pause() to remove those event listeners for those controller components. (The core aframe controller components would ideally be improved, but some of them are quite old and I'm working around them. pausing before removing can't hurt, but it does help.)

src/components/networked-hand-controls.js
kylebakerio: I call pause on the el before adding the controller component, so that I can manually bypass a part of the component before init() is fired. Then after I've updated the added controller component.

The part I'm bypassing is the part that sets up pose tracking for the controller. The controller components are designed with the assumption that they will only be used locally. I have surveyed the controllers, and found that there are basically two ways to get around this:

1) some controllers are better designed, and I can skip the pose detection from every being activated and just load the model separately.
2) some older ones cannot be run without pose detection being activated; for these, I remove pose detection right after it is added. (that's what the setInterval is for.)

src/components/networked-hand-controls.js
kylebakerio: I _think_ it's better to not do that in this case. `this.tick()` already has a job. This `setInterval()` will usually run once, but it absolutely will only run for maximum 25~ seconds if there's an error. It will always be removed. To add it to the tick, I'd need to set a flag in init, add a check on that flag on every tick at all times (even when this code isn't relevant--which, since there aren't many people using these old controllers, will usually be the case), and turn that flag on/off here.

The "optimization" of using tick() here seems to me like it would actually hurt, not help. Agreed?

kylebakerio: to be clear, ideally 
1. we wouldn't have this at all (which, we usually don't, if the component separates pose tracking and model injection and allows us bypass one without the other)
2. fallback: would be nice if we could 'listen' for when the tracking component gets added and remove it on-that-event, but there is no such event
3. we could do a 'setTimeout', but that would be gross; so, we do a 'setInterval' instead, on a short loop; while this setInterval almost always runs exactly one time in my testing, a setInterval provides a layer of safety in case something goes wrong.

kylebakerio: I did make some updates to this code. In testing, I was able to see that it usually only looped once even with a 1ms interval. I went ahead and switched it out to a `requestAnimationFrame` rather than do a 1ms interval, just because a 1ms interval if it does run seems like it could produce too much overhead if it were to run its full length, even if it only would be for a few seconds.

src/components/networked-hand-controls.js
kylebakerio: I see, wasn't aware of those auto-calls underneath, but that makes sense. I'll clean that up to just removeAttribute then and verify that it works as expected.

src/components/networked-hand-controls.js
```diff
@@ -336,29 +336,34 @@ AFRAME.registerComponent('networked-hand-controls', {
...
+        this.el.components['tracked-controls-webxr'].remove(); // in case support is added in the future
+        this.el.removeAttribute('tracked-controls-webxr'); // just some cleanup for inspector, etc.
+        this.el.play();
+      }
+      else {
+        requestAnimationFrame(disablePostTracking);
+      }
+    };
+
+    requestAnimationFrame(disablePostTracking);
```
vincentfretin, line 372: You probably meant Pose instead of Post? So disablePoseTracking?

You didn't have an issue with window.requestAnimationFrame? I remember there is an issue (rendering a pdf page on a canvas) with window.requestAnimationFrame not being executed in an XR session.
If it's working properly that's great, otherwise you can consider setting a flag on the component and execute disablePoseTracking if the flag is set in the tick method. You can find this pattern in the raycaster component with the dirty flag for example, there the refreshObjects is called in checkIntersections not in tick but it's the almost the same.

src/components/networked-hand-controls.js
```diff
@@ -336,29 +336,34 @@ AFRAME.registerComponent('networked-hand-controls', {
...
         } catch(e) {
           console.error("error removing controller component", e);
         }
         throw new Error("never found tracked-controls-webxr, cancelling interval that was waiting to remove it and removing controller components")
       }
-    },25)
+      else if (this.el.components['tracked-controls-webxr']) {
+        // there is no remove on this component in current version of A-Frame, so we kill it with pause and tick first.
+        this.el.components['tracked-controls-webxr'].pause();  // remove its event listeners
+        this.el.components['tracked-controls-webxr'].tick = () => {}; // remove on-tick pose tracking in the component
```
vincentfretin, line 362: I don't understand why patching tick is needed here. Did you encounter an issue that would have required this?
Removing the component calls pause and remove and the pause is a wrapped method that calls sceneEl.removeBehavior(component) that remove calling the tick method from the animation loop. From what I understand all that is done synchronously in the same tick.

vincentfretin: The interesting parts in aframe code
https://github.com/aframevr/aframe/blob/795edfd65841d54aa2b186a02f660ffee606b50c/src/core/component.js#L671-L672
https://github.com/aframevr/aframe/blob/795edfd65841d54aa2b186a02f660ffee606b50c/src/core/component.js#L763-L794
https://github.com/aframevr/aframe/blob/837023cca1a04b408b4acde4c09823e09dc58e6d/src/core/a-entity.js#L381-L405

vincentfretin: A behavior is a component that has a tick or tock method
https://github.com/aframevr/aframe/blob/795edfd65841d54aa2b186a02f660ffee606b50c/src/core/component.js#L753-L755
Here is the scene tick that execute each behavior tick if isPlaying
https://github.com/aframevr/aframe/blob/0dd84abe8fd25367d87da207ed9e4bbcd6f9d92d/src/core/scene/a-scene.js#L731-L740

src/components/networked-hand-controls.js
kylebakerio: > You probably meant Pose instead of Post? So disablePoseTracking?

yup!

> You didn't have an issue with window.requestAnimationFrame?

I can only test this by using an emulator. It seemed to work, but digging into it more, I see that it _may_ cause issues, rarely. I've updated the code to handle either case, and was able to document both window.requestAnimationFrame as well as xrSession.requestAnimationFrame being used with new code.

src/components/networked-hand-controls.js
kylebakerio: Just an artifact of the way I developed an approach to make this possible, along with me not knowing for sure what guarantees I had about this stuff. I've cleaned up this section of code, will commit soon.

src/components/networked-hand-controls.js
vincentfretin: Ok, looking at the whole code again, I think this is too hacky to check if a component has been set by regularly checking for it like that to then remove it. I think we better should check for the componentinitialized event and then call removeAttribute if this is the component name we're waiting for. You have an example in the animation component here:
https://github.com/aframevr/aframe/blob/837023cca1a04b408b4acde4c09823e09dc58e6d/src/components/animation.js#L452-L459

kylebakerio: Sure, I understand the feeling. I hadn't noticed that event before, that does look nice.

I just wrote up a version that uses it, unfortunately I also have to use a hack with it; if I don't do a 1 millisecond setTimeout, A-Frame seems to have a weird bug that locks it in an infinite loop if I remove in the event listener itself.

kylebakerio:
Through working on the quest 2 controller in parallel, I have realized that only some model changes are handled in `updateModel()`, but others are handled earlier that I'm missing and skipping replay on, in `onButtonChange()`. I am considering updating the code to be grabbing those events further up and relaying them, so that button movements are also handled, instead of just button mesh colorations.

Another option is to offer a pull request upstream to modify those controller components so that they truly do allow separate manipualtion of the controller without emitting an event. It could be doable without creating any breaking changes, just separation of concerns. But that's more work, anotehr pull request on another project, etc.

Kind of annoying. The architecture and flow of the events for controllers is more than a bit awkward.

kylebakerio:
This is particularly relevant for the quest 2 controllers, which I will be making a pull request to core for to enable full joystick and trigger button representation. I'd like those to be visible over the wire, and with the current design, that stuff would be missed.

I'll come back with more updates soon, need to run some experiments on that.

kylebakerio:
Alright, so:

1. I expanded the method I've used to grab joystick movements, mesh colors, and button mesh rotation/position updates.
2. I've tested it with the new not-released quest2 model component that support moving buttons; also, the vive controls component supports moving buttons, and that worked as well
3. Also tested with a few other models of older oculus controllers (to be clear: all except for quest 2 were only tested with emulated chrome extension)
4. This all works, correctly, in my experiments, for all the models I have tested recently (you can test the latest, including with experimental quest2 support, here).
5. However, knowing everything I know, if I were starting at the beginning again, I think I'd be tempted to just intercept and rebroadcast simplified versions of the actual base events from the `tracked-webxr-controls` component itself, for the cleanest implementation.
6. That said, there are some optimization benefits gained from doing it this way, so I'm not really motivated to rewrite this into that form--this seems to work quite well.

examples/js/meta-touch-controls.js
```diff
@@ -0,0 +1,507 @@
+/* global AFRAME, THREE */
```
kylebakerio: note: I added this file only so that I could test with it on glitch: https://naf-wip.glitch.me/tracked-controllers.html

it is removed in a later commit

examples/tracked-controllers.html
```diff
@@ -48,6 +50,10 @@
         debug: true;
         adapter: wseasyrtc;
     "
+      renderer="antialias: true;
+      colorManagement: true;
+      sortObjects: true;
+      physicallyCorrectLights: true;"
```
kylebakerio, line 56: these make seeing the controller buttons much easier, they are really hard to see without this

vincentfretin: okay

src/components/networked-hand-controls.js
```diff
@@ -153,38 +113,284 @@ AFRAME.registerComponent('networked-hand-controls', {
...
+
+        // so we first remove the controller component as best we can
+        if (this.el.components[this.data.controllerComponent]) {
+          // there is generally no remove() specified, but they do have pause(), which is close enough:
+          try {
+            this.el.components[this.data.controllerComponent].pause();
+            this.el.components[this.data.controllerComponent].remove();
+          } catch(e) {
+            console.error("NAF: possible problem while updating handModelStyle to controller", e)
+          }
```
vincentfretin, line 144: This try catch can be removed completely right? removeAttribute below is already doing the job.

src/components/networked-hand-controls.js
```diff
@@ -153,38 +113,284 @@ AFRAME.registerComponent('networked-hand-controls', {
...
+    });
+  },
+
+  updateHandMeshColor() {
+    const color = new THREE.Color(this.data.color);
+
+    // a different way to set color that we aren't using:
+    // this.getMesh().children[1].material.emissive = color;
+    // this.getMesh().children[1].material.color = color;
+    // this.getMesh().children[1].material.metalness = 1;
```
vincentfretin, line 226: Can you remove those comments? Also you don't need to create a THREE.Color instance here, `color.set(this.data.color)` should be fine with any string.

src/components/networked-hand-controls.js
```diff
@@ -153,38 +113,284 @@ AFRAME.registerComponent('networked-hand-controls', {
...
+    // this method is instead borrowed from the oculus-touch-controls component
+    this.getMesh().children[1].material.color.set(color);
+    this.el.sceneEl.systems.renderer.applyColorCorrection(this.getMesh().children[1].material.color);
+  },
+
+  controllerComponents: [
+    'magicleap-controls',
+    'vive-controls',
+    'oculus-touch-controls',
+    // 'meta-touch-controls', // TODO! REMOVE ME! ONLY FOR TESTING, DO NOT MERGE WITH THIS LINE UNCOMMENTED
```
vincentfretin, line 237: You can remove the comment.

src/components/networked-hand-controls.js
```diff
@@ -153,38 +113,284 @@ AFRAME.registerComponent('networked-hand-controls', {
...
+          this.el.components[this.data.controllerComponent].updateModel(...this.data.controllerModelUpdate)
+        }
+      }
     }
   },
 
   remove() {
-    this.el.removeObject3D('mesh');
+    this.el.removeObject3D(this.str.mesh);
+    this.removeEventListeners();
```
vincentfretin, line 191: Are you sure you want `this.removeEventListeners();` here? It seems wrong. `this.removeEventListeners();` is already in `pause()` for local player symmetrically with the `this.addEventListeners()` for the local player in `play()`

kylebakerio: yeah, that's correct. I hadn't touched A-Frame almost at all since last fall until the past month or so. Not sure if I forgot or never knew, but I've either learned or re-learned from you about the removeAttribute calling remove which calls pause pattern. when I started I assumed it was just incomplete. makes sense, though.

thanks for the review, I think everything I wrote would have worked fine either way, but I definitiely gained knowledge about writing better a-frame code.

vincentfretin (approved):

src/components/networked-hand-controls.js
vincentfretin: Thank you for your patience. I'm glad you learned some things from this review.

I also relearned yesterday that when adding a component, it runs `init` and then `update`!, then `play`.
See https://github.com/aframevr/aframe/blob/795edfd65841d54aa2b186a02f660ffee606b50c/src/core/component.js#L306-L336
I had a code in `update` that was replacing an outfit material for an avatar, and I saw it did the logic twice.
My code was this:
```
schema: {
    outfit: { default: "" },
},
update: function (oldData) {
    if (!oldData) return; // Normally always a {} if it's an object based component when adding the component locally, but can be undefined some times when NAF is creating it.
 
    if (oldData.outfit !== data.outfit) {
      this.loadOutfit();
    }
}
```
To fix the issue I just added an extra condition:
```
update: function (oldData) {
    if (!oldData) return; // Normally always a {} if it's an object based component when adding the component locally, but can be undefined some times when NAF is creating it.
 
    if (oldData.outfit && oldData.outfit !== data.outfit) {
      this.loadOutfit();
    }
}
```
That works because I know I'm setting the outfit when creating the component, I don't want a naked avatar. :)

vincentfretin:
I tested it on my Quest 1, it's working.
controller_quest1

kylebakerio:
My latest update for the oculus-touch-controls should also fix the quest 1 (and all non-quest-2 oculus controllers, I think) controllers having the 'all buttons light up when you touch one button' issue, and that'll instantly start working with this component at that time if/when it gets merged in.

PR #364, opened 2022-09-13
use better movement-controls version in examples

vincentfretin:
Use a version of movement-controls allowing your to move with left joystick, and moving camera with right joystick.


vincentfretin:
I updated the examples to a newer aframe-extras.controls.min.js version in https://github.com/networked-aframe/networked-aframe/commit/b092c6a5f41a67a532efeeaa283087b606666cc4

PR #368, opened 2022-09-17
sync quaternion instead of euler

vincentfretin:
Sync quaternion instead of the rotation euler to fix some rotation issues with the YXZ or XYZ euler order.
I didn't fix the tests yet.
@kylebakerio with that does it fix your rotation issues?

### DIFF
--- a/src/components/networked.js
+++ b/src/components/networked.js
@@ -409,7 +409,7 @@ AFRAME.registerComponent('networked', {
       if (this.networkUpdatePredicates[i](syncedComponentData) || fullSync) {
         componentsData = componentsData || {};
         if (componentName === 'rotation') {
-          // We compared the rotation, but we are sending actually the quaternion.
+          // We compared the rotation, but we actually send the quaternion.
           // JSON.stringify(new THREE.Quaternion(0,1,0,1)) gives '{"_x": 0,"_y": 1,"_z": 0,"_w": 1}' with "_" for each key
           // so we need to create an object for it to work properly. THREE.Vector3 doesn't have this issue.
           const q = componentElement.object3D.quaternion;
--- a/src/components/networked.js
+++ b/src/components/networked.js
@@ -1,7 +1,6 @@
 /* global AFRAME, NAF, THREE */
 var deepEqual = require('../DeepEquals');
 var InterpolationBuffer = require('buffered-interpolation');
-var DEG2RAD = THREE.MathUtils.DEG2RAD;
 var OBJECT3D_COMPONENTS = ['position', 'rotation', 'scale'];
 
 function defaultRequiresUpdate() {
@@ -139,8 +138,6 @@ AFRAME.registerComponent('networked', {
       el: this.el
     };
 
-    this.conversionEuler = new THREE.Euler();
-    this.conversionEuler.order = "YXZ";
     this.bufferInfos = [];
     this.bufferPosition = new THREE.Vector3();
     this.bufferQuaternion = new THREE.Quaternion();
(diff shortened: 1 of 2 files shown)


kylebakerio:
Yes! I compared it with master, and this branch does indeed fix it. 
Here's the problem on master:
[naf-rotation-problem.webm](https://user-images.githubusercontent.com/6391152/190879636-109fe828-e36c-4dc3-95da-188168adea71.webm)
Here's the issue fixed after this fix:
[naf-controller-fixed.webm](https://user-images.githubusercontent.com/6391152/190879684-4aafa0c6-9efa-4e9c-a71b-d6303122303e.webm)

src/components/networked.js
```diff
@@ -411,6 +408,21 @@ AFRAME.registerComponent('networked', {
       // Call networkUpdatePredicate first so that it can update any cached values in the event of a fullSync.
       if (this.networkUpdatePredicates[i](syncedComponentData) || fullSync) {
         componentsData = componentsData || {};
+        if (componentName === 'rotation') {
+          // We compared the rotation, but we are sending actually the quaternion.
```
kylebakerio, line 412: "but we are sending actually" -> "but we actually send" or, "but we are actuallying sending"

src/components/networked.js
```diff
@@ -560,8 +572,8 @@ AFRAME.registerComponent('networked', {
         buffer.setPosition(this.bufferPosition.set(data.x, data.y, data.z));
         return;
       case 'rotation':
-        this.conversionEuler.set(DEG2RAD * data.x, DEG2RAD * data.y, DEG2RAD * data.z);
-        buffer.setQuaternion(this.bufferQuaternion.setFromEuler(this.conversionEuler));
+        this.bufferQuaternion.copy(data);
```
kylebakerio, line 575: Nice, this looks much cleaner now.

vincentfretin:
Out of curiosity, did you try before adding rotation component on the entities of left-hand-default-template and right-hand-default-template templates, as well on my-tracked-left-hand and my-tracked-right-hand entities so this set the YXZ order? Didn't this fix the issue too?
Even if it fixed the issue, putting an empty rotation component just to set the same rotation order as the hard coded YXZ in naf is counter-intuitive, I agree.

kylebakerio:
> Out of curiosity, did you try before adding rotation component on the entities of left-hand-default-template and right-hand-default-template templates, as well on my-tracked-left-hand and my-tracked-right-hand entities so this set the YXZ order? Didn't this fix the issue too? Even if it fixed the issue, putting an empty rotation component just to set the same rotation order as the hard coded YXZ in naf is counter-intuitive, I agree.

I just hadn't gotten around to trying.

vincentfretin:
Just adding `templateInner.setAttribute('rotation', '0 0 0'); // to set the YXZ order` in `addHandTemplate`
fixes the issue as well.
The avatar has the `look-controls` component on it so it's setting the YXZ order because it has `rotation` component as dependency, but on the receiving side in avatar-template we have the default XYZ because we don't specify the rotation component on the template, but even if the order mismatch we don't seem to have the issue.
FYI Hubs has a `set-yxz-order` component they use on rig, camera, hands and media to fix this issue actually, see
https://github.com/mozilla/hubs/search?q=set-yxz-order
and in their new NetworkedTransform they are syncing the quaternion now.

vincentfretin:
I pushed https://github.com/networked-aframe/networked-aframe/commit/0d46822205f98a584a19bf16bc99fd3b2f3687a4 on master to immediately fix the issue. 
I'll work further on this PR, fixing the tests obviously, but also I want to try modifying the serialization to use an array instead of an object with x,y,z,w keys for position, rotation, scale and using on the receiving side `Vector3/Quaternion.fromArray(data)` instead of `Vector3/Quaternion.copy(data)`

src/components/networked.js
vincentfretin: Thanks, I changed to "but we actually send".

vincentfretin:
That would be a non issue in aframe 1.6.0.
aframe master now set `object3D.rotation.order = 'YXZ';` even when the rotation component is not set.
https://github.com/aframevr/aframe/pull/5513

PR #401, opened 2023-03-12
Allow networked entities to become children of a non-networked parent

Agupta00:
## What

Here we add the ability for networked entities to become childeren of non-networked entities (Ie, allow them to end up somewhere other than the root a-scene) 

## Why

For example, imagie each client has a local platform that is _not_ networked, but we want entities to become childeren off. Right now there is no way from my understanding to do this within the general API. 

## Before

You can see the other client ends up below the platform ( because they are not a child of the platform)

https://user-images.githubusercontent.com/12666226/224524346-ba26c6d1-ce85-47c6-b855-40c01aac1ea2.mp4

## After

The other client ends up on the platform (because it is a child of it)

https://user-images.githubusercontent.com/12666226/224524353-ae67b9b6-147e-43a2-b717-6b30d6acdade.mp4

### DIFF
--- a/examples/basic.html
+++ b/examples/basic.html
@@ -100,6 +100,8 @@
       <a-entity environment="preset:starry;groundColor:#000000;"></a-entity>
       <a-entity light="type:ambient;intensity:0.5"></a-entity>
 
+      <a-box id="platform" color="tomato" depth="10" height=".5" width="10" position="0 10 0">
+
       <!--   Here we declare only the local user's avatar, which we then broadcast to other users     -->
       <!--   The 'spawn-in-circle' component will set the position and rotation of #rig;
              because this entity also has the networked component, and position and rotation are tracked by default,
@@ -108,7 +110,7 @@
              including all applicable child elements. However, because we don't need to see our own avatar, we use the
              `attachTemplateToLocal:false` option. This makes our local copies invisible on our machine, but visible on everyone else's.
       -->
-      <a-entity id="rig" movement-controls="fly:true;" spawn-in-circle="radius:3" networked="template:#rig-template;">
+      <a-entity id="rig" movement-controls="fly:true;" spawn-in-circle="radius:3" networked="template:#rig-template; attachToParentId:platform;">
         <!-- Here we add the camera. Adding the camera within a 'rig' is standard practice.
          We set the camera to head height for e.g. computer users, but otherwise never touch it again; if the user enters VR,
          its rotation and position will be updated by the headset in VR. If we need to touch the user's position
@@ -125,6 +127,8 @@
         >
         </a-entity>
       </a-entity>
+
+      </a-box>
     </a-scene>
 
     <script>
--- a/src/NetworkEntities.js
+++ b/src/NetworkEntities.js
@@ -88,6 +88,7 @@ class NetworkEntities {
 
   receiveFirstUpdateFromEntity(entityData) {
     var parent = entityData.parent;
+    var attachToParentId = entityData.attachToParentId;
     var networkId = entityData.networkId;
 
     var parentNotCreatedYet = parent && !this.hasEntity(parent);
@@ -96,7 +97,7 @@ class NetworkEntities {
     } else {
       var remoteEntity = this.createRemoteEntity(entityData);
       this.createAndAppendChildren(networkId, remoteEntity);
-      this.addEntityToPage(remoteEntity, parent);
+      this.addEntityToPage(remoteEntity, parent, attachToParentId);
     }
   }
 
@@ -121,9 +122,11 @@ class NetworkEntities {
     }
   }
 
-  addEntityToPage(entity, parentId) {
+  addEntityToPage(entity, parentId, attachToParentId) {
     if (this.hasEntity(parentId)) {
       this.addEntityToParent(entity, parentId);
+    } else if (attachToParentId) {
+      this.addEntityToParentId(entity, attachToParentId)
     } else {
       this.addEntityToSceneRoot(entity);
     }
@@ -131,6 +134,16 @@ class NetworkEntities {
 
   addEntityToParent(entity, parentId) {
     this.entities[parentId].appendChild(entity);
+  }
+
+  addEntityToParentId(el, attachToParentId) {
+    var parent = document.querySelector(`#${attachToParentId}`);
+    if (!parent) {
+      NAF.log.warn('Tried to add entity to parent, but parent with id', attachToParentId,
+        'is not in the scene');
+      return;
+    }
+    parent.appendChild(el);
   }
 
   addEntityToSceneRoot(el) {
--- a/src/components/networked.js
+++ b/src/components/networked.js
@@ -117,6 +117,7 @@ AFRAME.registerComponent('networked', {
   schema: {
     template: {default: ''},
     attachTemplateToLocal: { default: true },
+    attachToParentId: { default: '' },
     persistent: { default: false },
 
     networkId: {default: ''},
@@ -427,6 +428,7 @@ AFRAME.registerComponent('networked', {
     syncData.template = data.template;
     syncData.persistent = data.persistent;
     syncData.parent = this.getParentId();
+    syncData.attachToParentId = data.attachToParentId;
     syncData.components = components;
     syncData.isFirstSync = !!isFirstSync;
     return syncData;


src/components/networked.js
```diff
@@ -427,6 +428,7 @@ AFRAME.registerComponent('networked', {
     syncData.template = data.template;
     syncData.persistent = data.persistent;
     syncData.parent = this.getParentId();
+    syncData.attachToParentId = data.attachToParentId;
```
vincentfretin, line 431: I see we already have `syncData.parent` here, so adding a new field to sync a parent id seems redundant. I think it would be better to modify `getParentId` and `initNetworkParent` here, and keep the `addEntityToPage(entity, parentId)` signature.

examples/basic.html
```diff
@@ -108,7 +111,8 @@
              including all applicable child elements. However, because we don't need to see our own avatar, we use the
              `attachTemplateToLocal:false` option. This makes our local copies invisible on our machine, but visible on everyone else's.
       -->
-      <a-entity id="rig" movement-controls="fly:true;" spawn-in-circle="radius:3" networked="template:#rig-template;">
+      <a-entity id="rig" movement-controls="fly:true;" spawn-in-circle="radius:3" networked="template:#rig-template; attachToParentId:#platform;">
```
vincentfretin, line 114: Instead of modifying the basic example. Can you create a new example where you absolutely need your changes.
For this modified example, instead of `attachToParentId:#platform;"` you can really just use `position="0 10 0"` on rig and it will be absolutely the same.

src/components/networked.js
Agupta00: If we wanted to do that I think we would need to add smt like a prefix for non-networked parents in `getParentId()` like `nonNetworked:${ParentId}` 

since over [here](https://github.com/networked-aframe/networked-aframe/blob/f1c62b41a4a13c6d4d6f425eaf14c81ab90e44ff/src/NetworkEntities.js#L93): where we do `    var parentNotCreatedYet = parent && !this.hasEntity(parent);`, . 

there we would have to parse to check if the parentId on `syncData` is a networked parent or a non-networked parent. (We have to do this because we have two cases, 1) The networked parent is not in the scene yet and we add to childCache, or 2) the parent is not a networked component so we can just add it to the scene right now)

examples/basic.html
Agupta00: Yeah sure, did you have an idea of what would be good? Perhaps I can change it so the client can manually set the hight of the platform local to his own computer. Then I think you would absolutely need this change. Does that work?

vincentfretin: I'm asking you because I don't know which use case those changes would really be useful. If I merge changes, I want an example where the changes are necessary and the example can't be achieved in another way.
Having the height of the platform configurable doesn't change anything, you change both position y on rig and platform and that would do it without the changes.
Didn't you have a real use case for those changes?
Your modified example seems the wrong use case to me. What about a second platform, where the avatars can go, you will need to change the parent of your avatar, how would it work? Generally you would not put your avatar rig as a parent of another entity but use a root entity, you would then update the avatar rig according to where the avatar is. This is how it works with the two different navmesh components we have.

Agupta00: Apologies in advance if I misunderstood anything.., here is why I think my example works. 

Example: Player A's platform's y=10, Player B's platform is at y=20. 

On Player B's local client, how would you ensure that Player A ends up on my platform? If you set Player A rig entity position to y=20, then the next time Player A moves, the rig's position would get updated back to the original y height since its rig is networked entity right? If it is still unclear I can make the example and perhaps you can try it. 

_(side note): I tried setting the position of a networked entity (from another client) and it did not update, maybe there is logic that prevents this anyways?_


**Real Life Example**

Here is a video of a real use case, I can attach the scene code if it helps, but in summary, each client has a sort of "platform" whose position is local the client. In the video the "platform" is the green outline, which is tracked and changed locally on the client. A different phone viewing the same target can have a different position for the "platform".

The thing I care about syncing is the robot's positions, but only relative to the "platform". You can think of the "platform" as the root of my networked scene.  

https://user-images.githubusercontent.com/12666226/225457407-78c2ef4e-bd81-43c6-a822-1ccbafa03e51.mp4


```
  <a-scene>
    <a-entity id="platform"> //local client can set this to anything. The position here only makes sense to the client so it should not be networked.
      <a-entity id="avatar" networked=""template:#avatar-template> // The robot's position in the scene should make sense relative to other players in the scene as well. 
      </a-entity>
    </a-entity>
  </a-scene>
```

vincentfretin: For your side note, you need to call `NAF.utils.takeOwnership(yourEntity)` to be the owner of it so the changes propagate to the other users.
Ok if you have a different platform height for each user, your example indeed is fine, and of course you don't want it networked.
What I wanted to understand is why you need a platform height different for each user.
From your video I really don't know what I'm seeing.
But I can see that it can probably be useful if you set the platform height corresponding to an anchor on a table in AR for example, although you could just set the position on a-scene I think.
It would be interesting to have an example with one user on desktop and another with Chrome Android with AR on the phone moving a character like your video.

vincentfretin: Yes please go ahead.
I don't know if you have an ARCore compatible Android phone. If you do, you can look at creating an experience in AR to place the platform on a table, see https://github.com/AdaRoseCannon/aframe-xr-boilerplate for an example. Chrome Android is only implementing hittest, that's enough to have the correct height of the table. Out of scope here, but I think you can now detect a plane on Meta browser on Quest, so probably you can know the table boundaries and set the width and height of the platform accordingly too.

Agupta00:
Updated readme with video of the new example html as well.

vincentfretin:
Just to let you know, I didn't forget your PR. I was busy with other things. I'll take a deeper look at your changes and do some tests myself when I'll get the chance.

Agupta00:
> Just to let you know, I didn't forget your PR. I was busy with other things. I'll take a deeper look at your changes and do some tests myself when I'll get the chance.

Np! And thanks for your reply :)

travaglinit:
Is this PR/fork the recommended approach for nesting networked entities beneath a non-networked parent? I am looking to implement something similar where different clients might set different positions for the non-networked parent that has networked children

vincentfretin:
@travaglinit You can test the changes and give feedback on it. I didn't took the time yet to review it again and experiment on my side, sorry. I'll probably do that when I'll play with the new anchored component from aframe 1.5.0 and Quest 3.

examples/index.html
```diff
@@ -53,6 +53,10 @@ <h1>Networked-Aframe Examples</h1>
       <a href="basic-events.html">Basic with events</a>
       <span>Example of listening to and logging NAF events</span>
     </div>
+    <div>
+      <a href="non-networked-paraent.html">Non Networked Parent</a>
```
travaglinit, line 57: Should be "non-networked-parent.html"

examples/index.html
```diff
@@ -53,6 +53,10 @@ <h1>Networked-Aframe Examples</h1>
...
+      <span>Demonstrates how to attach networked entities as childeren of non-networked parents</span>
```
travaglinit, line 58: Should be "entities as children"

examples/non-networked-paraent.html
```diff
@@ -0,0 +1,188 @@
...
+             `attachTemplateToLocal:false` option. This makes our local copies invisible on our machine, but visible on everyone else's.
+      -->
+        <a-entity
+          id="rig"
+          movement-controls="fly:true;"
+          spawn-in-circle="radius:3"
+          networked="template:#rig-template; allowNonNetworkedParent:true;"
+        >
+          <!-- We add allowNonNetworkedParent because we want networked entities to end up as a child of platform entity not the root a-scene.
+          -->
```
travaglinit, line 160: I like that we've added this documentation as comments. I would recommend also adding additional information visible in the UI of this example explaining what this example is and how to use it without needing to inspect the source code. See the "Persistent spheres" example which includes the header at the top as an example.

src/NetworkEntities.js
```diff
@@ -122,8 +132,11 @@ class NetworkEntities {
   }
 
   addEntityToPage(entity, parentId) {
+    var nonNetworkedParent = document.querySelector(`#${parentId}`)
     if (this.hasEntity(parentId)) {
       this.addEntityToParent(entity, parentId);
+    } else if (nonNetworkedParent) {
```
travaglinit, line 138: I would recommend setting line 135 to just `var nonNetworkedParent;` and then the assignment for nonNetworkedParent could instead happen on this line as:

`else if ((nonNetworkedParent = document.querySelector(`#${parentId}`)))`

That way we can avoid running the querySelector call if we enter the first if condition.

travaglinit (approved):
Added a couple comments, but nothing that would lead to breaking changes, so approving this PR.

travaglinit:
@vincentfretin sounds good. I reviewed the PR and added a couple of small comments, but nothing major so this PR should be good. I'll send a PR later in a different fork that includes the changes I mentioned. 

I've tested the example @Agupta00 added, and it works as expected. I also re-tested all of the other examples and those are still working as expected, and the unit tests are passing.

travaglinit:
> @vincentfretin sounds good. I reviewed the PR and added a couple of small comments, but nothing major so this PR should be good. I'll send a PR later in a different fork that includes the changes I mentioned.
> 
> I've tested the example @Agupta00 added, and it works as expected. I also re-tested all of the other examples and those are still working as expected, and the unit tests are passing.

I've sent the PR from my fork which addresses the PR comments:

https://github.com/networked-aframe/networked-aframe/pull/440

PR #405, opened 2023-04-12
Clarifying hand models versus hand tracking

Utopiah:
The example can be confusing leading users to believe that hand tracking is supporting while it is, for now at least, not.


vincentfretin:
Thanks!

PR #418, opened 2023-06-12
Fix low volume on Safari iOS 16.5

vincentfretin:
See https://bugs.webkit.org/show_bug.cgi?id=236219 and workaround described in comment https://bugs.webkit.org/show_bug.cgi?id=218012#c37 to fix low volume on Safari iOS 16.5 compared to Chrome iOS.


vincentfretin:
Apparently the same changes that I did in janus adapter don't work here with the easyrtc adapter.
You can try
```
<script src="https://cdn.jsdelivr.net/gh/networked-aframe/networked-aframe@80705d568cef8c1be814df38c83271a3693ac1da/dist/networked-aframe.min.js"></script>
```

Actually even on the janus adapter, the workaround don't work anymore this morning, that's crazy.

vincentfretin:
Ahmed on discord is telling me on this easyrtc example
https://bald-scrawny-protoceratops.glitch.me/demos/demo_audio_only.html
that the sound is working perfectly fine and the sound is coming from speakers, loud and perfect.

And that on https://naf-examples.glitch.me/basic-audio.html with positional audio or not, the sound is coming only on earpiece (or both earpiece and speaker?), the sound is really low.
It's hard for me to test and found workaround because I don't have an iPhone. On iPad, the only Apple device I own, there is no earpiece, so I don't have the issue at all.

vincentfretin:
You can sponsor $50 to [vincentfretin](https://github.com/sponsors/vincentfretin) to expedite this fix.

Issue #394, opened 2023-01-30
didn't manage to get hand tracking working


Utopiah:
disentangling from https://github.com/networked-aframe/networked-aframe/issues/381

Could someone please share a working example with both URL and video recording of a real session (no emulator)?

kylebakerio:
[hand-tracking-quest2.webm](https://user-images.githubusercontent.com/6391152/221280067-b5006577-c0ee-4456-8eea-d8921cf633be.webm)

Sure, motion in this vid is done in quest 2, recording is from desktop. url: https://naf-examples.glitch.me/tracked-controllers.html

Did it just now. Would love to know what's going on with your setup, but vincent and I have never been able to replicate your issues unfortunately, would need new info.

kylebakerio:
Ah! 

This component is *not* compatible with the actual hand tracking API. I do realize the naming is confusing, but that's because it's inhereting the naming from the a-frame component that predates the hand-tracking API. In other words, this is a networked version of A-Frame's [hand-controls](https://aframe.io/docs/1.4.0/components/hand-controls.html) component.

Specifically, the hand controls component generates "hands" with gestures that match certain button click configurations. This component networks those gestures, and further supports a 'controller' model as an option that shows the actual buttons and joystick clicks instead of the faux-hand and faux-gestures.

In both cases, though, yes, controllers are required.

I've wanted to make a component that networks the skeleton pieces from the hand-tracking API--I think that would be magical and great--but I never got around to it. 

But I'm glad that clear it up. In this video demo, one hand is rendered as a controller, and one is rendered as a hand, just to show both options of display--but both are indeed controllers.

Utopiah:
How would one go to start with just hand position? I'm thinking a sphere again would be a good start. It's not as precise as a hand model with position, rotation, and all joints, but I believe it's enough for some fun behaviors, including waving, pointing to objects, etc.

Utopiah:
Working example https://twitter.com/utopiah/status/1711375479335227530 if others are interested. One can't "just" use a template with the hand tracked entity as it does not have a post but rather traverse through it to get the wrist (and optionally other joints) then make an invisible entity then network that one. An indirect solution with some overhead but seems to work.

PR #445, opened 2023-12-14
Add networked hand tracking component

avdynut:
Hi, thanks for the networked component.
We're adding it to the [XR# CommunityToolkit](https://github.com/XRSharp/XRSharp.CommunityToolkit). So multi-user experience would quickly be available for C# developers.

It would be really nice to have hand-tracking shared between users. 
Here is a simple component that adds `networked` attribute for every hand joint. Only `modelStyle: dots` is supported.

Please check example project: https://glitch.com/edit/#!/naf-hand-tracking

Related issue #394

### DIFF
--- a/src/components/networked-hand-tracking.js
+++ b/src/components/networked-hand-tracking.js
@@ -1,0 +1,34 @@
+﻿/* global AFRAME */
+
+AFRAME.registerComponent('networked-hand-tracking', {
+    dependencies: ['hand-tracking-controls'],
+
+    schema: {
+        options: { default: 'template:#hand-joint-template; attachTemplateToLocal:false' }
+    },
+
+    init: function () {
+        var el = this.el;
+        this.trackingControls = el.components['hand-tracking-controls'];
+
+        this.onExitVR = AFRAME.utils.bind(this.onExitVR, this);
+        el.sceneEl.addEventListener('exit-vr', this.onExitVR);
+    },
+
+    tick: function () {
+        var jointEls = this.trackingControls.jointEls;
+
+        for (var i = 0; i < jointEls.length; i++) {
+            jointEls[i].setAttribute('networked', this.data.options);
+        }
+    },
+
+    onExitVR: function () {
+        var jointEls = this.trackingControls.jointEls;
+        this.trackingControls.jointEls = [];
+
+        for (var i = 0; i < jointEls.length; i++) {
+            jointEls[i].removeAttribute('networked');
+        }
+    },
+});
--- a/src/index.js
+++ b/src/index.js
@@ -7,3 +7,4 @@ require('./components/networked');
 require('./components/networked-audio-source');
 require('./components/networked-video-source');
 require('./components/networked-hand-controls');
+require('./components/networked-hand-tracking');


src/components/networked-hand-tracking.js
```diff
@@ -0,0 +1,34 @@
+﻿/* global AFRAME */
+
+AFRAME.registerComponent('networked-hand-tracking', {
+    dependencies: ['hand-tracking-controls'],
+
+    schema: {
+        options: { default: 'template:#hand-joint-template; attachTemplateToLocal:false' }
```
vincentfretin, line 7: I don't see the hand-joint-template in your PR. What did you use for html template?
Answering this myself looking at your glitch:
```html
<template id="hand-joint-template">
  <a-sphere radius="0.01"></a-sphere>
</template>
```

What about the networked schema? Default networked schema is only syncing position and rotation, did you add scale? I see hand-tracking-controls is setting scale on the joints https://github.com/aframevr/aframe/blob/e0c4ec6d2d07db426d27d76b32ec7aba7d9f5e60/src/components/hand-tracking-controls.js#L250
You probably want to create your own schema that sync position rotation scale with a correct requiresNetworkUpdate function to minimize networked updates, see
https://github.com/networked-aframe/networked-aframe#tracked-controllers-w-synced-gestures

```html
          <a-entity id="left" hand-tracking-controls="hand:left; modelStyle:dots;" networked-hand-tracking="{options: 'template:#hand-joint-template; attachTemplateToLocal:false'}"></a-entity>
```
I didn't know we could use an object like that as attribute value, this is really working?

src/components/networked-hand-tracking.js
```diff
@@ -0,0 +1,34 @@
...
+        var jointEls = this.trackingControls.jointEls;
+
+        for (var i = 0; i < jointEls.length; i++) {
+            jointEls[i].setAttribute('networked', this.data.options);
+        }
+    },
+
+    onExitVR: function () {
+        var jointEls = this.trackingControls.jointEls;
+        this.trackingControls.jointEls = [];
```
vincentfretin, line 28: I don't think this should be the responsibility of your component to empty the array of another component.
Why this line? It doesn't seem to be needed to me.

src/components/networked-hand-tracking.js
```diff
@@ -0,0 +1,34 @@
...
+        this.onExitVR = AFRAME.utils.bind(this.onExitVR, this);
+        el.sceneEl.addEventListener('exit-vr', this.onExitVR);
+    },
+
+    tick: function () {
+        var jointEls = this.trackingControls.jointEls;
+
+        for (var i = 0; i < jointEls.length; i++) {
+            jointEls[i].setAttribute('networked', this.data.options);
+        }
```
vincentfretin, line 23: Doing a setAttribute in tick for all joints are not efficient. You should really do it only once in an enter-vr listener I think.

vincentfretin (commented):
Thank you for the PR. When we looked at how to network hands last time with @Utopiah I didn't pay attention that the dots model was creating aframe entities, so indeed we can networked those easily.

I left some comments. I didn't test that myself though. @Utopiah you may be interested to testing it and give feedback.

If we want it to be merged, we'll need some documentation in the README and an example.

But we probably want a component that works with the hand models as well.
Performance wise, the implementation here is using 25 networked entities for a hand, that may work good enough for two users or three in the room, but I guess we could have perf issues quickly with more users.
What would be far better is having a component syncing directly the two Float32Array jointPoses and jointRadii, naf doesn't support binary transfer, so it will still be needed to serialize those to an Array to serialize in json unfortunately.

Utopiah:
@avdynut do you have a live Glitch to test?

Using the dots model isn't ideal and yet networking it all this way is too costly, yet I'd argue at this point something inefficient yet functional with known limits, e.g 2 participants only, is better than nothing, as is currently the case.

I'd suggest to accept the PR with minimal change and clear documentation and keep the improved version (hand model and Float32Array) for a subsequent PR.

UpTheDesign:
How about the following considerations to make it easier/lighter to network so much data:
1) Quantize the data
2) Minimize which joints you network
3) Only send update if movement exceeds a threshold
4) Differential/delta encoding

Then perhaps once per second or two, send a full "keyframe" without the above optimizations to catch for any glitches from 2, 3 or 4.

Additionally, consider an adaptive approach in which the extent of optimization depends on the amount of hands being tracked. So a 2 person event will track every joint with full accuracy etc, while a 50 person event will be closer to just tracking spheres...

PR #448, opened 2023-12-29
Expose InterpolationBuffer on NAF global

vincentfretin:
Can be useful to have it exposed so we can use it instead of adding another version of it in a project.
See my comment https://github.com/msub2/aframe-exokit-avatars/issues/8#issuecomment-1816916724


vincentfretin:
I used this change to sync the Head bone quaternion with interpolation here:
https://twitter.com/vincentfretin/status/1742904482801856557

Issue #450, opened 2024-01-20
networked-hand-controls issues with aframe-v1.5.0


bruce90j:
The networked-hand-controls component doesn't work properly with aframe-v1.5.0:
1. Hand color is not updated (example: `networked-hand-controls="hand:left;color:gold;"`).
    To make it work replace:
    https://github.com/networked-aframe/networked-aframe/blob/0f8452331d1d62fed9a76aaf4ab29961c1cc064e/src/components/networked-hand-controls.js#L219-L220
    with:
    ```
    const handColor = this.data.color;
    newMesh.traverse(function (object) {
        if (!object.isMesh) { return; }
        object.material.color = new THREE.Color(handColor);
    });
    ```

2. Hand controller does not work if controllerComponent is not specified (example: `networked-hand-controls="hand:right;handModelStyle:controller;"`).
    Remove `'gearvr-controls', 'daydream-controls'` from the `controllerComponents` list:

    https://github.com/networked-aframe/networked-aframe/blob/0f8452331d1d62fed9a76aaf4ab29961c1cc064e/src/components/networked-hand-controls.js#L235-L249

vincentfretin:
Can you please create a PR?

vincentfretin:
PR is #451

PR #451, opened 2024-01-22
fix networked-hand-controls issues with aframe-v1.5.0

bruce90j:
The networked-hand-controls component doesn't work properly with aframe-v1.5.0:
1. Hand color is not updated (example: `networked-hand-controls="hand:left;color:gold;"`).
    To make it work replace:
    https://github.com/networked-aframe/networked-aframe/blob/0f8452331d1d62fed9a76aaf4ab29961c1cc064e/src/components/networked-hand-controls.js#L219-L220
    with:
    ```
    const handColor = this.data.color;
    newMesh.traverse(function (object) {
        if (!object.isMesh) { return; }
        object.material.color = new THREE.Color(handColor);
    });
    ```

2. Hand controller does not work if controllerComponent is not specified (example: `networked-hand-controls="hand:right;handModelStyle:controller;"`).
    Remove `'gearvr-controls', 'daydream-controls'` from the `controllerComponents` list:

    https://github.com/networked-aframe/networked-aframe/blob/0f8452331d1d62fed9a76aaf4ab29961c1cc064e/src/components/networked-hand-controls.js#L235-L249

### DIFF
--- a/src/components/networked-hand-controls.js
+++ b/src/components/networked-hand-controls.js
@@ -203,7 +203,6 @@ AFRAME.registerComponent('networked-hand-controls', {
     this.loader = new THREE.GLTFLoader();
     this.loader.setCrossOrigin('anonymous');
     const handmodelUrl = this.MODEL_BASE + this.MODEL_NAMES[this.data.handModelStyle + this.data.hand.charAt(0).toUpperCase() + this.data.hand.slice(1)];
-    
     this.loader.load(this.data.customHandModelURL || handmodelUrl, gltf => {
       const newMesh = gltf.scene.children[0];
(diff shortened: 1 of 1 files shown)


vincentfretin:
Thanks @bruce90j, I also did the change in `updateHandMeshColor` in https://github.com/networked-aframe/networked-aframe/commit/5ccb8af576cb11c3bfc68eb65f65597a24783954
I tested with WebXR Chrome extension and the following command in the console:
`document.querySelector('[networked-hand-controls]').setAttribute('networked-hand-controls', 'color:green')`

vincentfretin:
This closes #450

Issue #243, opened 2021-02-24
Implement keepalive message in socketio adapter


vincentfretin:
You may think the socketio adapter is stable/reliable when just using the node server, this may be true, but it is not the case when you use it behind nginx, see #200
The socketio adapter (and also the webrtc adapter) in this repository lacks any sort of keepalive message, so if there is no component updates going through the socket for 30s (it depends on the keepalive_timeout param in nginx), the WebSocket will just close.

Please use the easyrtc/wseasyrtc adapters with open-easyrtc for production behind nginx, open-easyrtc is sending a stillAlive message every 20s by default.

I'm open to any PR that implement a keepalive message like it's done in open-easyrtc or janus adapter via the minijanus library if you want inspiration. I won't implement it myself though.

vincentfretin:
You can also implement something yourself in you app to send regularly a component update like `periodic-full-syncs` component or your avatar, see https://github.com/mozilla/hubs/blob/5fc48ef25336057ef7fb985eac5b2fcb3937f567/src/components/periodic-full-syncs.js

vincentfretin:
I see a ping pong message every 25s with socketio. Closing this.

vincentfretin:
Also from https://socket.io/docs/v4/using-multiple-nodes/ 

> [!CAUTION]
> The value of nginx's [proxy_read_timeout](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout) (60 seconds by default) must be bigger than Socket.IO's [pingInterval + pingTimeout](https://socket.io/docs/v4/server-options/#pinginterval) (45 seconds by default), else nginx will forcefully close the connection if no data is sent after the given delay and the client will get a "transport close" error.

Issue #242, opened 2021-02-23
Client doesn't properly reconnect when the server restart


vincentfretin:
Tested with adapters easyrtc wseasyrtc socketio and webrtc
with two avatars in the basic.html room.
If I restart the nodejs server, it seems there is a reconnection, I have some logs saying data channel open, but the avatars doesn't reappear.
This is only works as expect with the janus adapter.

For the webrtc adapter, there is also this issue #204

vincentfretin:
#200 is probably the same issue.

vincentfretin:
Example with easyrtc adapter:
![Capture d’écran de 2021-02-23 14-00-16](https://user-images.githubusercontent.com/112249/108847086-8d143280-75df-11eb-95f5-4ac45609766b.png)

So it doesn't do a resync of the entities.

vincentfretin:
It seems to me that socketio websocket detects the socket closed and reconnect but the adapter is not aware of it.
I know that the janus adapter that is using low level WebSocket uses a close event listener to call `this.reconnect()` that does a proper `this.disconnect()` and then `this.connect()` if the socket closed abnormally (`event.code !== 1000`)
https://github.com/mozilla/naf-janus-adapter/blob/2ce5e2c4fbcda573f175f22062a50d8ead0680ad/src/index.js#L260-L271

Anyone with socketio knowledge wants to fix this?

hthetiot:
I think this should work on easyrtc side. I will investigate later what is missing in the adapter to support automatic reconnected and call peers.

vincentfretin:
For socketio and webrtc adapters, this is fixed in #479 
The issue remains for (ws)easyrtc adapters, a similar fix is needed.

PR #458, opened 2024-03-20
Implement room instancing in socketio-server.js

vincentfretin:
Use case: experience with 3000 users, using room instancing by batch of 50 users, no audio, no video, only websocket. We are interested only in avatar movements.

For 50 in a room, you probably want to reduce the update rate to 1 per second instead of 15 by setting `NAF.options.updateRate = 1` ([doc](https://github.com/networked-aframe/networked-aframe?tab=readme-ov-file#options))

If you test it inside the networked-aframe repo, be sure to use `npm run dev-socketio` and setting `networked-scene="adapter:socketio"` in the basic.html or basic-chat.html examples.

Be aware there are pending issues with the socketio adapter:
- #242
- #243

That's a quick implementation I tested only with `maxOccupantsInRoom = 3` instead of 50.
I didn't test the load. Someone has to create a script to test 3000 connections, creating the websocket, sending avatar positions every second, do that for a random duration about 30 seconds to 3 min for each connection and close the websocket connection to see how the node process supports the load.

If someone wants to go further, some things to be tested for cpu/memory/latency:
- Use a more performance framework, replace express by [fastify](https://github.com/fastify/fastify/) (nodejs only) or [hono](https://hono.dev/) (on [deno](https://deno.com/) or [bun](https://bun.sh/)), or [h3](https://github.com/unjs/h3) (on node or deno or bun)
- Replace socketio by a more efficient websocket implementation (ws on nodejs?), modify the socketio adapter to use WebSocket directly instead of socketio.
- Or rewrite the server part entirely in [golang](https://go.dev) or [rust](https://www.rust-lang.org)/[tokio](https://tokio.rs) or [elixir/phoenix](https://www.phoenixframework.org)

### DIFF
--- a/server/socketio-server.js
+++ b/server/socketio-server.js
@@ -9,6 +9,9 @@ process.title = "networked-aframe-server";
 // Get port or default to 8080
 const port = process.env.PORT || 8080;
 
+// Threshold for instancing a room
+const maxOccupantsInRoom = 50;
+
 // Setup and configure Express http server.
 const app = express();
 
@@ -32,7 +35,7 @@ app.use(express.static(path.resolve(__dirname, "..", "examples")));
 const webServer = http.createServer(app);
 const io = require("socket.io")(webServer);
 
-const rooms = {};
+const rooms = new Map();
 
 io.on("connection", socket => {
   console.log("user connected", socket.id);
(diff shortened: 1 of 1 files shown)


Utopiah:
I'd be curious about metrics here, i.e how many users (10, 100, 1000...) managed on which hardware (RPi Zero, RPi5, i9...) and what is the bottleneck.

vincentfretin:
If you need perf, I'll probably rewriting server/socketio-server.js with 
https://github.com/uNetworking/uWebSockets.js
You can still use socketio lib on the client side I guess.

vincentfretin:
test done on my machine, 32GB of memory, 8 cores
50 users on http://localhost:8080/basic.html

# socketio adapter
a room of 50 users is about 30% cpu for networked-aframe-server, 0.3% memory, smooth movement, my 8 cores were at 78% for 50 chrome windows
5 rooms of 10 users is 10% cpu (I changed the maxOccupantsInRoom variable to 10 in server/socketio-server.js)

# wseasyrtc adapter
a room of 50 users is about 80% cpu for networked-aframe-server, 1.7% memory, some choppy movements, I don't know if it's server side or client side, probably both, because I was testing both clients and server on the same machine, my 8 cores were at 95%
20 users in a room is smooth
30 users smooth at the beginning but we lost a few during the test

# easyrtc adapter (so WebRTC datachannel instead of WebSocket for data, no audio)
a room of 50 users, very choppy movement
networked-aframe-server 60%, increasing memory +0.1% every second, memory not freed at the end

# janus adapter

20 users in a room with audio, cpu 84%, memory 0.3%, some choppy movements
transports over datachannel
EVENT_LOOPS=4 with adapter.setEventLoops(4)
MESSAGE_THREADS=1
MAX_ROOM_SIZE=20
ALLOW_LOOP_INDICATION=true

20 users in a room with audio, cpu 78%, memory 0.3%, some choppy movements, my 8 cores were at 98% for 20 chrome windows
transports over websocket
EVENT_LOOPS=4 without adapter.setEventLoops(4)
MESSAGE_THREADS=1
MAX_ROOM_SIZE=20
ALLOW_LOOP_INDICATION=false

20 users in a room with audio, cpu 63%, memory 0.3%, some choppy movements with networked-audio-source="positional: true", smooth movements without networked-audio-source or with networked-audio-source="positional: false", my 8 cores were at 68% for 20 chrome windows. With networked-audio-source="positional: true", my 8 cores were at 88%
transports over websocket
EVENT_LOOPS=4 with adapter.setEventLoops(4)
MESSAGE_THREADS=1
MAX_ROOM_SIZE=20
ALLOW_LOOP_INDICATION=true

Chrome profiling gives sometimes linearRampToValueAtTime 4.26 ms with positional audio, so hearing the positional audio is really the bottleneck on the browser.

vincentfretin:
@Utopiah for now maximum is 70 users in a room with socketio server with socketio 4.8.1 running with node or bun on a Raspberry Pi 400 (ARM Cortex-A72 Quad-core 1.8 GHz)
https://x.com/vincentfretin/status/1865678789906161967

PR #509, opened 2025-12-06
allow using exist entitys for networked

arpu:


src/NetworkEntities.js
```diff
@@ -16,20 +15,34 @@ class NetworkEntities {
...
 
-    var networkId = entityData.networkId;
-    var el = NAF.schemas.getCachedTemplate(entityData.template);
+    try {
+      const exists = document.getElementById(networkId);
+      const el = exists || NAF.schemas.getCachedTemplate(template);
+
+      if (!el) {
+        throw new Error(`Failed to create element for template: ${template}`);
+      }
```
vincentfretin: `NAF.schemas.getCachedTemplate(template)` already log an error if there is no template, so just return here maybe?

src/NetworkEntities.js
```diff
@@ -16,20 +15,34 @@ class NetworkEntities {
...
 
-    this.registerEntity(networkId, el);
+      this.registerEntity(networkId, el);
 
-    return el;
+      return exists ? null : el;
+    } catch (error) {
+      console.error('Error creating remote entity:', error);
+      return null;
+    }
```
vincentfretin: The try catch is not needed if you don't throw your own error.

src/NetworkEntities.js
```diff
@@ -16,20 +15,34 @@ class NetworkEntities {
 
   createRemoteEntity(entityData) {
     NAF.log.write('Creating remote entity', entityData);
+    const { networkId, template, components } = entityData;
 
-    var networkId = entityData.networkId;
-    var el = NAF.schemas.getCachedTemplate(entityData.template);
+    try {
+      const exists = document.getElementById(networkId);
```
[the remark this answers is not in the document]
vincentfretin: Add the log `NAF.log.write('Creating remote entity', entityData);` we already have abvove only if existingEl is null, and add another log if it already exists.

src/components/networked.js
```diff
@@ -209,6 +209,38 @@ AFRAME.registerComponent('networked', {
...
+      const isMine = this.isMine();
+
+      if (isMine && !wasMine) { // Gained ownership
+        this.removeLerp();
+        this.syncAll(undefined, true);
+        this.onOwnershipGainedEvent.oldOwner = oldData.owner;
+        this.el.emit(this.OWNERSHIP_GAINED, this.onOwnershipGainedEvent);
+      } else if (!isMine && wasMine) { // Lost ownership
+        this.onOwnershipLostEvent.newOwner = data.owner;
+        this.el.emit(this.OWNERSHIP_LOST, this.onOwnershipLostEvent);
```
vincentfretin: Not sure about all the changes here. We already have the some OWNERSHIP_LOST event emitted in networkUpdate. Is is emitted twice now?

src/components/networked.js
```diff
@@ -209,6 +209,38 @@ AFRAME.registerComponent('networked', {
...
+    const data = this.data;
+
+    // If owner has changed
+    if (oldData.owner !== data.owner) {
+      const wasMine = oldData.owner === NAF.clientId;
+      const isMine = this.isMine();
+
+      if (isMine && !wasMine) { // Gained ownership
+        this.removeLerp();
+        this.syncAll(undefined, true);
```
vincentfretin: Why do you always full sync now with the second param isFirstSync to true.

vincentfretin (commented):
That's a change we talked in January right?
Can you please give me again the scenario you have that is not covered by the persistent system? If it's possible to add an example with your case that will help, and probably additional unit test cases to cover the changes.

PR #514, opened 2026-02-21
Fix easyrtc and wseasyrtc adapters to work with open-easyrtc >= 2.1.5

vincentfretin:
Update open-easyrtc to 2.1.7, fix easyrtc and wseasyrtc adapters to call `connect` before `joinRoom`, that's required since 2.1.5.

The change of behavior of the `joinRoom` function for a patch version was reported in https://github.com/open-easyrtc/open-easyrtc/issues/142

Claude Code:
Here's a summary of the changes:

Root cause: `setRoom()` was calling `easyrtc.joinRoom()` immediately, but NAF calls `setRoom()` before `connect()` (see NetworkConnection.js:37 vs :57). Since open-easyrtc 2.1.5, `joinRoom()` now throws if `webSocket` isn't set yet.

Fix in both `WsEasyRtcAdapter.js` and `EasyRtcAdapter.js`:

1. `setRoom()` — removed the `easyrtc.joinRoom()` call, now just stores the room name
2. `connect()` — moved `joinRoom()` to after `easyrtc.connect()` succeeds, using its success/failure callbacks to then proceed with `connectSuccess`/`connectFailure`

This matches the pattern used in all open-easyrtc demo examples: set up listeners first, connect, then join rooms only after the websocket is established.

Vincent:
That's what we do already in `naf-socketio-adapter.js` and `naf-uws-adapter.js` so the changes are coherent with all adapters now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "buffered-interpolation": "Infinitelee/buffered-interpolation#5bb18421ebf2bf11664645cdc7a15bd77ee2156b",
         "express": "^4.17.3",
-        "open-easyrtc": "^2.1.0",
+        "open-easyrtc": "^2.1.7",
         "socket.io": "^4.8.1"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


vincentfretin:
Fix included in networked-aframe 0.14.1

